PR #2, opened 2015-11-13
get articles before rendering

nickhsine:
@hcchien


hcchien:
It seems not working.

PR #58, opened 2015-12-16
use className and style object for react component

kaddopur:
@nickhsine 

- use `className` and `style` object
- remove `text-align` because it's already defined in `.nav_logo` css


tom76kimo:
:+1:

Issue #60, opened 2015-12-16
apply eslint and editorconfig


kaddopur:
@nickhsine 

Maybe we could start from presets like
- https://www.npmjs.com/package/eslint-config-airbnb
- https://www.npmjs.com/package/eslint-config-rackt

Editorconfig is a very useful tool to make the code style consistent
https://github.com/rackt/redux/blob/master/.editorconfig
```yml
# EditorConfig helps developers define and maintain
# consistent coding styles between different editors and IDEs.

root = true

[*]
end_of_line = lf
charset = utf-8
trim_trailing_whitespace = true
insert_final_newline = true
indent_style = space
indent_size = 2
```

nickhsine:
@kaddopur :100:  :+1: 
thanks for your advice.
Becasue we are focusing bug fix now, will adopt eslint soon.

However, if you want to give us some PRs, we will totally be appreciated :)

yutin1987:
++ :100:

kaddopur:
I could do the PR and fix the styles :smile: 
But don't know the `indent` size for this project
Or let me just use `eslint-config-rackt` preset to start with?

nickhsine:
Thanks Jason for helping!
We will try to fit the eslint after days.

PR #63, opened 2015-12-16
apply eslint and editorconfig

kaddopur:
@nickhsine  #60 

I am not sure this style preset works for everyone, but If you have any concern please just let me know.
Let me try the style on `/server` first and if the rules works, I will fix other files gradually.


nickhsine:
Thank you so much!

PR #71, opened 2015-12-16
Bump react-slick version to support react 0.14

jessy1092:
To resolve 'getDOMNode is deprecated' warning.

On react-slick `0.9.2`. Carousel would autoplay when user unfocus page. So I also hack `visibilitychange` to stop autoplay carousel when page is not focus or visible

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "react-inline-css": "^2.0.0",
     "react-redux": "^2.1.2",
     "react-router": "^1.0.0-rc4",
-    "react-slick": "0.6.6",
+    "react-slick": "0.9.2",
     "react-transform-catch-errors": "^1.0.0",
     "react-transform-hmr": "^1.0.1",
     "redux": "^3.0.0",
--- a/src/components/TopNews.js
+++ b/src/components/TopNews.js
@@ -44,8 +44,8 @@ export default class TopNews extends Component {
                 }
                 const image = imageComposer(a, device);
                 return (
+                    <div key={a.id}>
                     <a
-                        key={a.id}
                         href={(a.slug) ? "https://www.twreporter.org/a/" + a.slug : a.storyLink}
                         className="topnewsimage-wrap"
                         style={{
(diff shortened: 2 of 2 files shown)


nickhsine:
Really appreciate your work.
But last time we use the latest version of react-slick, we encountered some bugs. Hence, we downgraded the package version.
We will try your PR tomorrow again.

Thanks :)

nickhsine:
I have verified the results.
It seems pretty good; the bug disappeared.
But there is one problem:
the carousel won't move automatically after onVisibilityChange emitted 
``` 
    this.setState({
        autoplay: true
    });
````
seems not working though.

jessy1092:
@nickhsine 
After I watch the `react-slick` source code, I found It didn't use `componentWillReceiveProps` so it would lose timer after I change the `autoPlay` property and never come back.

It must click the button or arrow to set timer again.

I create the PR https://github.com/akiran/react-slick/pull/199 for support `autoplayPause` property. It can be more flexible for user. :)

nickhsine:
@jessy1092 
It's cool.
In the meanwhile, maybe we can skip ```onVisibilityChange``` event handler first.
We can add it later after ```react-slick``` accept the PR.

jessy1092:
@nickhsine 
I think stop autoplay is better than infinite loop? 

In my test, the carousel would show the empty vision if unfocus page for a long time. 
BTW, the currently `react-slick`(0.6.6) also won't move automatically after focus page. 
Behavior is the same. XD

nickhsine:
@jessy1092 
It would be great if we have ```autoplayPause```,
but I just think we can upgrade ```react-slick``` version first.
The event handler could be added later.

However, if you want to wait ```react-slick```,
we can wait for it.

jessy1092:
@nickhsine 
OK, I remove the `onVisibilityChange` event. Please help to review. Thanks :)

nickhsine:
:+1:  :100: :shipit:

PR #82, opened 2015-12-18
Load more functionality for feature items

nickhsine:
Because the data structure of api responses are different (fetching from ```API_HOST/tags``` and ```API_HOST/article```),
there are two business logic in reducer(```LOADED_MULTI_TAGGED_ARTICLES_SUCCESS```,
```LOADED_ARTICLES_SUCCESS```) to handle them.
@hcchien 
@kaddopur


hcchien:
ok

PR #83, opened 2015-12-18
add Category component

kidwm:
I also made it responsive by adopting this svg background and used in topnews and featurebox with different size.

Maybe you can remove the `hexagon.png` after use this component everywhere.


kidwm:
BTW, this also make it possible that category name have more than two characters.

hcchien:
great!

PR #90, opened 2015-12-21
choose image to show on client side

nickhsine:
@hcchien


hcchien:
please update the component/Tag.js at the same time.

PR #100, opened 2015-12-22
remove nodemon

kidwm:
and use server.js as middleware directly with browserSync.


nickhsine:
:100:

kidwm:
Sorry, there are some problems still remain to solve.

nickhsine:
Do we need to rollback this commit?

kidwm:
@nickhsine Not needed. I'll address them in next PR.

PR #103, opened 2015-12-23
Navigation bar and category page

nickhsine:
@hcchien


src/components/Tags.js
```diff
@@ -19,15 +35,11 @@ export default class Tags extends Component {
     if (articles && articles.length > 0) {
       return (
         <div className="category-items">
-          <div className="category-name">
-            <Category>{cat_display}</Category>
-          </div>
+        <cat cat_display={cat_display}/>
```
nickhsine: @hcchien where is ```cat``` ?

hcchien: it still doesn't work. so I just remove it first.

hcchien:
we will need to wait for the design spec before merge.

hcchien:
ok, merge it.

PR #132, opened 2016-01-04
make logo fit in 414px width(below) device

nickhsine:
@hcchien


hcchien:
so it works from the device width less than 414px

nickhsine:
YES

PR #151, opened 2016-02-03
Add API server and make all the XHR requests pass through the API layer

nickhsine:
runtime transpilation for es6/7

add eslint ignore file

move browsersync-server to bin folder and add global variable

add webpack global variable

Update package.json
1. add dependencies
2. add custom script

Update server/server.js
1. add error handling for http proxy
2. add eslint condition

Add api layer on FE server
1. change redux action
2. add corresponding actions of api layer

Add bin/api.js and api/api.js
delete unused files

move constants variable into constants.js

@hcchien @garfieldduck


src/middleware/api.js
```diff
@@ -7,6 +9,15 @@ import _ from 'lodash'
 
 export const CALL_API = Symbol('CALL_API')
 
+function formatUrl(path) {
+  if (__SERVER__) {
+    // Prepend host and port of the API server to the path.
+    return 'http://' + config.apiHost + ':' + config.apiPort + path
```
hcchien: would you include the protocol to the variable?

api/constants.js
```diff
@@ -0,0 +1,5 @@
+let config = {
+  API: 'https://www.twreporter.org/api'
+}
+
+export default config
```
hcchien: why don't you put them into the config.js?

nickhsine: ok

src/middleware/api.js
nickhsine: on server side, we only use ```http``` protocol.

PR #159, opened 2016-03-02
modify src/conf/category-tag-mapping-table.js

nickhsine:
@hcchien


nickhsine:
tag of category page for 觀點 should be ```cat:評論```

PR #166, opened 2016-04-26
add article action/reducer and unit test

nickhsine:
Since we are going to use [new API layer](https://github.com/twreporter/tr-projects-rest/tree/keystone),
I just modify the ```redux state``` [data structure](https://github.com/twreporter/twreporter-react#redux-state). Please help to review this ```redux state``` data structure.

This article action is to fetch a single article from new API layer.

@hcchien @garfieldduck


src/actions/article.js
```diff
@@ -0,0 +1,72 @@
...
+    slug,
+    article,
+    receivedAt: Date.now()
+  }
+}
+
+function fetchArticle(slug) {
+  return dispatch => {
+    dispatch(requestArticle(slug))
+    return fetch(`http://${config.apiHost}:${config.apiPort}/posts/${slug}`)
```
hcchien: protocal should be included in the config.

PR #176, opened 2016-05-10
Fix bugs

nickhsine:
1. Webpack development server listening on wrong port, should be 3001, not 30001
2. remove ```encodeURIComponent(nextUrl)```

@garfieldduck @hcchien


garfieldduck:
nice!

PR #184, opened 2016-05-14
response 404 page while article is not found

nickhsine:
- add NotFoundError, InternalServerError and FatalError
- create NotFoundError and InternalServerError when fetching article
  occurs error
- add fatalError in redux state
- code refactor for device reducer

@garfieldduck


garfieldduck:
:100:

PR #199, opened 2016-05-25
Bugs fix

nickhsine:
1. extract bootstrap styles from inline to file to fix the problem of rendering react components before loading bootstrap css.

2. use lodash to ```get``` and ```set``` from and into object.


garfieldduck:
👍 🚀 ☀️

PR #201, opened 2016-05-27
Upgrade lodash

roderickhsiao:
and avoid direct require full lodash library to reduce client side byte size.

https://github.com/lodash/lodash/releases/tag/4.0.0


hcchien:
💯

PR #221, opened 2016-06-14
Load articles

nickhsine:
1. article schema change, set idAttribute: 'id', not 'slug'
2. add slugToIds, which is a Map, in state
3. rename articles.test.js to group-articles.test.js and unit test
updates
4. remove redundant code in src/reduers/article.js
5. add action and reducer for articles
6. fetch related articles after component mount
7. 	use single schema(author) to represent writers, engineers, designers and photographers


src/actions/articles.js
```diff
@@ -0,0 +1,91 @@
...
+function receiveArticles(response, ids) {
+  return {
+    type: types.FETCH_ARTICLES_SUCCESS,
+    ids,
+    response,
+    receivedAt: Date.now()
+  }
+}
+
+function buildQueryURL(ids = []) {
```
hcchien: Maybe you should change the function name if it just allows the query by ids.

src/actions/articles.js
```diff
@@ -0,0 +1,91 @@
...
+  let query = {}
+  let where = {
+    '_id': {
+      '$in': ids
+    }
+  }
+  query.where = JSON.stringify(where)
+  query.sort = '-publishedDate'
```
nickhsine: what do you mean? by both ?

src/actions/articles.js
```diff
@@ -0,0 +1,91 @@
...
+      let normalized = normalize(camelizedJson.items, arrayOf(articleSchema))
+      dispatch(receiveArticles(normalized, ids))
+    }, (error) => {
+      return dispatch(failToReceiveArticles(ids, error))
+    })
+  }
+}
+
+
+export function fetchArticlesIfNeeded(ids) {
```
hcchien: change the function name to fetchArticlesByIdsIfNeeded?

src/actions/articles.js
nickhsine: query not only contains ```ids```, but also contains ```sort``` and ```embedded```

PR #223, opened 2016-06-16
Aside page navigation

nickhsine:
@garfieldduck @hcchien


src/components/article/PageNavigation.js
```diff
@@ -0,0 +1,110 @@
+'use strict'
+import { Link } from 'react-router'
+import _ from 'lodash'
+import classNames from 'classnames'
+import React, { Component } from 'react'
+import ReactDOM from 'react-dom'
+import styles from './PageNavigation.scss'
```
hcchien: sorting? (upper case, lower case)

hcchien:
👍

src/containers/Article.js
```diff
@@ -188,7 +188,14 @@ export default class Article extends Component {
...
         </div>
-
+        <ArticleComponents.PageNavigation
+          article={ _.get(article, [ 'relateds', 0 ])}
+          navigate="next"
+        />
+        <ArticleComponents.PageNavigation
+          article={_.get(article, [ 'relateds', 1 ])}
+          navigate="previous"
+        />
```
garfieldduck: Do we need further discussion on how to let users go back to the previous article they've just clicked? Sry I just found NYT can let user go back while the first previous article might be selected from its backend

nickhsine: Sorry, it's just a temporary solution for showing next and previous articles.
It should be discussed later after the designer and HC come into office.

garfieldduck:
LGTM 👍 👍 👍

PR #225, opened 2016-06-20
Update InfoBox and Paragraph components

nickhsine:
code change because of this [pr](https://github.com/twreporter/keystone/pull/58)
However, link style is not tuned yet.
Will do it after spec is final.

@garfieldduck


garfieldduck:
LGTM 👍

PR #233, opened 2016-06-23
Update serveral files

nickhsine:
- transpile dependent files as well by babel
- concat css files and move the concated css file to src/styles

@garfieldduck


garfieldduck:
:+1:  :+1:  :+1:

PR #237, opened 2016-06-24
Style tunning according to new design

nickhsine:
Only mobile version, waiting for desktop and tablet spec.
@garfieldduck

### DIFF
--- a/server/views/index.ejs
+++ b/server/views/index.ejs
@@ -1,6 +1,5 @@
 <!DOCTYPE html>
-<!-- bootstrap default font-size: 10px, change the default font-size by style here -->
-<html lang="zh-Hant-TW" style="font-size:18px">
+<html lang="zh-Hant-TW">
   <head>
       <title>報導者The Reporter</title>
       <meta http-equiv="content-type" content="text/html; charset=utf-8" />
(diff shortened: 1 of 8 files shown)


server/views/index.ejs
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
-<html lang="zh-Hant-TW">
+<!-- bootstrap default font-size: 10px, change the default font-size by style here -->
+<html lang="zh-Hant-TW" style="font-size:18px">
```
garfieldduck: put default font size in bootstrap.overrides.scss

src/components/article/InfoBox.scss
```diff
@@ -1,16 +1,55 @@
+@import '../../../node_modules/bootstrap-sass/assets/stylesheets/bootstrap/_variables.scss';
+@import '../../themes/font-mixins.scss';
+@import '../../themes/media-queries.scss';
+
 .infoBoxContainer {
-  padding: 25px 50px 25px 50px;
+  $color: #262626;
```
garfieldduck: put color as a common variable

src/components/article/InfoBox.scss
```diff
@@ -1,16 +1,55 @@
...
+  $logo-width: 23px;
+  $logo-height: 24px;
```
garfieldduck: use the rem mixin?

PR #238, opened 2016-06-26
mini-refine for revamp branch for component key

aqutw:
mini-refine for revamp branch for component key

### DIFF
--- a/src/components/Daily.js
+++ b/src/components/Daily.js
@@ -23,14 +23,14 @@ export default class Daily extends Component {
         </div>
         <div className="daily-itemlistwrapprt">
           <ul className="daily-itemlist">
-          { _.map(dailyTop, (a) => {
+          { _.map(dailyTop, (a, idx) => {
             const pubDate = date2yyyymmdd(a.publishedDate, '.')
             if (a.state == 'published') {
               let imageSet = imageComposer(a)
               let thumbnail = imageSet.mobileImage
               let url = '/a/' + a.slug
               return (
-                <li className="daily-item" key={a._id}>
+                <li className="daily-item" key={a.id || idx}>
                   <a className="clearfix" href={url}>
                     <div className="daily-image" >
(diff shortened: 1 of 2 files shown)


src/components/Daily.js
```diff
@@ -23,14 +23,14 @@ export default class Daily extends Component {
...
-          { _.map(dailyTop, (a) => {
+          { _.map(dailyTop, (a, idx) => {
             const pubDate = date2yyyymmdd(a.publishedDate, '.')
             if (a.state == 'published') {
               let imageSet = imageComposer(a)
               let thumbnail = imageSet.mobileImage
               let url = '/a/' + a.slug
               return (
-                <li className="daily-item" key={a._id}>
+                <li className="daily-item" key={idx}>
```
nickhsine: The new data structure we have right now should have ```id``` property.
Could you use ```{a.id || idx}``` here?

nickhsine:
Thank you for your contribution.

👍

src/components/Daily.js
aqutw: 1.
Although `{a.id || idx}` is also a way, 
I first prefer that decoupling Frontend logic from Backend value when they are not related.

2.
Secondly, I prefer `{a.id}` because it make sure all DOM use value from Backend.

3.
I don't prefer use `{a.id || idx}`:
For the value of key, this way might be ambiguity.

PS:
Sometimes, we need share the same value between Frontend & Backend parts together, 
but here we are not.

nickhsine: ```idx``` is just a fallback.
If you don't want to use fallback, I could still accept it.

However, I don't like to use ```{idx}``` alone, it might cause some problems while re-rendering.

aqutw: Agree~

Then `{a.id}` should be the best.

aqutw:
I updated revampB via your suggestion, sorry for late.

PR #242, opened 2016-07-01
Update Audio.js

nickhsine:
According to this [PR](https://github.com/twreporter/keystone/pull/61), we will get ```parsed image object```, so there is no need to parse it again.
@hcchien


hcchien:
👍

PR #246, opened 2016-07-08
Fix bug. Calculate right width of slideshow container

nickhsine:
move ```LazyLoad``` as the children of ```<div className={styles['ss-slider']} ref={i => this._slideshow = i}>```.
If ```LazyLoad```  is the parent of slideshow component, the width of slideshow will be mis-calculated.

@garfieldduck


garfieldduck:
👍 ☀️ :shipit:

PR #250, opened 2016-07-20
Extract font-weight into common-variables

garfieldduck:
- Extract font-weight into common-variables
- Fix a bug in PublishDate
- Center content on desktop


nickhsine:
👍 LGTM

PR #261, opened 2016-08-02
Display topic box if the article is part of a topic

garfieldduck:
Display topic box if the article is part of a topic


hcchien:
👍

PR #265, opened 2016-08-09
Deprecate react-article-components. Build twreporter-react npm pkg.

nickhsine:
### Major update:
- Build twreporter-react npm pkg by
``` npm run build-pkg```
All js, css and scss files under ```src``` folder will be transpiled by babel into ```lib``` folder.

- Deprecate react-article-components. Use twreporter-react instead.

### Other updates
- move some files, used only by ourselves, under src/lib/ to src/utils.
- remove unused files.

@garfieldduck


garfieldduck:
Awesome! :+1: :rocket:  :shipit:

PR #268, opened 2016-08-10
Revamp related articles in the bottom according to the new spec

garfieldduck:
- adjust styles
- add load more function

TODO:
- combine topic if the article is in the one of the topic


nickhsine:
💯 🚀 🚢 
學到了幾招 css 用法，感謝！

PR #290, opened 2016-08-15
Fix hover style of the global css, link styles of Paragraph, width of some article components

garfieldduck:
- fix hover styles of the global css
- link styles of Paragraph
- fix problems in displaying common styles of some article components


nickhsine:
💯 🚀 🚢

PR #299, opened 2016-08-17
Adjust tag and index page navbar

garfieldduck:
- show categories when user scrolls down on desktop
- fix tags listing page alignment problem
- refine tag listing style
- replace footer's div with html5 tag
- add topic title in topic's index page
- change title to h1 for SEO

@nickhsine


nickhsine:
💯 🚢

PR #301, opened 2016-08-18
Show topic title on navbar when scrolled down, Fix photography NavBar style

garfieldduck:
- show topic title on navbar when scrolled down
- fix photography NavBar style
- replace 500 font-weight with 400 in some CSS
- change some items name in the menu

@nickhsine


nickhsine:
👍 💯 🚢

PR #302, opened 2016-08-18
Several Enhancement

nickhsine:
Show url to fetch in the redux log
Do not fetch related articles while articleId is not provided
New line in text inside the QuoteBy component

@garfieldduck


garfieldduck:
:+1: :100: :shipit:

PR #306, opened 2016-08-18
Show placeholders when loading articles

garfieldduck:
- Show image and text placeholders when loading articles
- Fix photography list item styling

@nickhsine


nickhsine:
NICE 👍 🐐

PR #309, opened 2016-08-19
Use feature articles as fallback when there is no related articles

nickhsine:
1. Use feature articles as fallback when there is no related articles
2. Update BlockAlignmentWrapper.scss

@garfieldduck


garfieldduck:
awesome :+1:  :sunny:  :shipit:

PR #310, opened 2016-08-19
Set photography style for the article page

garfieldduck:
- set photography style for the article page
- remove separate text color settings for each components in Article


nickhsine:
💯 👍 ⛳ :shipit:

PR #315, opened 2016-08-22
Make Embedded align center even if it is too small

nickhsine:
[Reference: Get iframe height and width while editing.](https://github.com/twreporter/keystone/pull/103) 
And then we can use width and height to align the iframe well.
@garfieldduck


garfieldduck:
:100: :bike: :shipit:

PR #320, opened 2016-08-24
Replace storage url string to match our ATS remap rule

nickhsine:
@garfieldduck


garfieldduck:
LGTM :+1:  :+1:  :+1:

PR #323, opened 2016-08-24
set maxwidth for lazy-loaded images

garfieldduck:
### Bug Fix:
set maxwidth for lazy-loaded images

@nickhsine


nickhsine:
LGTM 💯

PR #331, opened 2016-08-29
Fine tune index page styling, Leave enough padding for header when page changed

garfieldduck:
- fix top margin of the footer
- adjust characters limits of the article header
- fix styles of the 'more article' button
- fix the display problem of indicator on the Home slider

@nickhsine


nickhsine:
💯 🚢

PR #332, opened 2016-08-29
Do not use feature articles as fallback on those footer article page

nickhsine:
@garfieldduck


garfieldduck:
:100:  :+1:  :shipit:

PR #335, opened 2016-08-30
Use <a> tag rather than Link of react-router when the style of article is interactive

nickhsine:
@garfieldduck


garfieldduck:
:100: :ship: :shipit:

PR #336, opened 2016-08-30
Use lodash to get author image, Add article SEO tag, Global css styles tuning

garfieldduck:
- Use lodash to get author image
- Add article SEO tag
- Global css styles tuning

@nickhsine


nickhsine:
🚢 :shipit:

PR #339, opened 2016-08-30
fix topic item overflow, refine various header stylings

garfieldduck:
- fix topic item overflow
- refine top spacing as page switches
- hide read progress bar as page switches
- refine topic text display on header


nickhsine:
💯 👍

PR #357, opened 2016-09-08
Image css files optimizations, Remove unnecessary event listener, Refine css of Audio

garfieldduck:
- add proper css animations to control buttons on the slideshows
- optimize css files related to image display
- remove unnecessary event listener
- refine css of Audio

@nickhsine


nickhsine:
🚢 :shipit:

PR #367, opened 2016-09-14
fix image lazyload effect, add google amp, extend api timeout

garfieldduck:
* fix image lazyload effect
* add google amp
* extend api timeout

@nickhsine


nickhsine:
👍 💯 🚢

PR #421, opened 2016-11-16
[Update] Improve article lazyload effect on desktop devices

garfieldduck:
- Improve article lazyload effect on desktop devices
- Change the outer shadow of the article loading placeholder

@nickhsine


nickhsine:
👍 💯

PR #438, opened 2016-12-08
[Update] Update api/actions/loadArticles.js

nickhsine:
- Not to get images of authors to decrease API calls
- Only get a single full article for now


garfieldduck:
Awesome! 💯 :shipit:

PR #439, opened 2016-12-15
[Feature] Topic landing page and Promotion banner. 

nickhsine:
### [Add] Add src/containers/TopicLandingPage.js 
- Add landing page of topic
- Fetch whole topic data if needed
- Render all related articles in the topic by using Cards component
- Render LeadingVideo or LeadingImage in the first screen
- Rendere description and team description in the middle of the topic page

### [Update] Update api/actions/loadArticles.js
- Not to get images of authors to decrease API calls
- Only get a single full article for now

### [Add] LeadingVideo which works on all devices. 
1. Autoplay and play inline with sound on Desktop device
2. Autoplay and play inline without sound on mobile device, except for
those devices whose iOS versions are below 10.
3. Render leading image rather than video on iOS9 below.

### [Add] Add PromotionBanner component
- render PromotionBanner both in the index page and article page

@garfieldduck


garfieldduck:
LGTM 👍 👍 👍

PR #461, opened 2017-01-04
[Update] fix image placeholder color for photography

hanyulo:
fix the bug of image placeholder of photography page

### DIFF
--- a/src/components/article/Image.js
+++ b/src/components/article/Image.js
@@ -141,8 +141,10 @@ class Image extends FitwidthMixin(Component) {
       </div>
     )
 
+    const { isPhotography } = this.context
+
     return (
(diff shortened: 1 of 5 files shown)


src/components/article/Image.js
```diff
@@ -141,8 +141,10 @@ class Image extends FitwidthMixin(Component) {
       </div>
     )
 
+    let isPhotography = this.context.isPhotography
```
garfieldduck: const {isPhotography} = this.context

garfieldduck:
👍 👍 👍

src/components/article/Image.js
hanyulo: mission accomplished!

PR #463, opened 2017-01-08
[Update] Update caching.

nickhsine:
- Add health check endpoint.
- Increase expire time of redis record.
- Increase max-age of Cache-Control directive.

@garfieldduck


garfieldduck:
:100:

PR #464, opened 2017-01-16
[Update] Update package.json. Replace forever with pm2 and give the --max_old_space_size directive

nickhsine:
## Problem on Production : JavaScript heap out of memory
**Temporary solution** 
1. replace forever with pm2 ( to see logs easily)
2. Add --max_old_space_size to increase memory size to 2GB.

The right solution needs more logs,  so currently use this change to monitor the production. 

@garfieldduck

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   },
   "betterScripts": {
     "start-prod": {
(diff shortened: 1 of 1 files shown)


garfieldduck:
💯

PR #469, opened 2017-01-18
[Update] Update processes.json. Add max_memory_restart and increase the --max_old_space_size

nickhsine:
**Temporary Solution for preventing server out of memory**

Increase v8 max_old_space_size and restart server by pm2 when memory usage is over the threshold.


garfieldduck:
💯 🚢

PR #474, opened 2017-01-23
[Update] Enable print media query and print the article well

nickhsine:
**[Problem Description]** <br>
Currently our article page does not support web print (網頁列印), so you may see wrong layout of printing page.

**[Solution]**

1. Change media attribute in the css script tag
2. Use bootstrap print media query
3. Do not show navbar, video, audio, related bottoms and footer component while printing.

@garfieldduck @YuCJ @hanyulo


garfieldduck:
LGTM 👍

PR #488, opened 2017-02-09
[Bug] Show correct Header on article page

nickhsine:
**BUG DESCRIPTION**

Header contains subNavMenu on article page.

**RESULT**

There is no subNavMenu on article page.

@garfieldduck


garfieldduck:
💯 :shipit: 🚢

PR #499, opened 2017-02-15
[AuthorPage] Move functions from action to reducer

YuCJ:
- Move state formatting from action creator to reducer
- Add author collection items per page from 3 to 5

### DIFF
--- a/src/reducers/author.js
+++ b/src/reducers/author.js
@@ -5,13 +5,11 @@ import { REQUEST_PAGE_START_FROM } from '../constants/author-page'
 
 import get from 'lodash/get'
 import isArray from 'lodash/isArray'
-import merge from 'lodash/merge'
 import uniq from 'lodash/uniq'
 
 const _ = {
   get,
   isArray,
-  merge,
   uniq
 }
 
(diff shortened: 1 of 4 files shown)


src/reducers/author.js
```diff
@@ -31,9 +34,25 @@ export const author = (state = initialStates, action = {}) => {
...
+        isFetching: false,
+        [authorId]: {
+          collectIndexList: _.get(response, 'result', []),
+          currentPage,
+          totalResults,
+          receivedAt,
+          isFinish: (currentPage - REQUEST_PAGE_START_FROM + 1 >= totalPages)
+        }
+      }
+      return _.mergeWith({}, state, saveToState, concatCollectIndexList)
```
nickhsine: All you need is to merge the `collectIndexList`, so I think you can directly concat the collectIndexList and uniq it without using `mergeWith` function.

YuCJ: The difference between using `mergeWith` and only concacting the `collectIndexList` is that  using `mergeWith` will keep the author(A) collection list and pagination info in state when user go to other author(B)'s page. It will not need to refetch collections list of authorA if it has been saved in state.

YuCJ:
- Concat collectIndexList before merge
- remove useless `authorId` property in state

nickhsine:
👍 💯

PR #538, opened 2017-03-24
[Update] Code updates according to pagespeed advice

nickhsine:
- defer script tag to avoid blocking
- load typekit asynchronously
- not to lazy load leading image

### DIFF
--- a/src/components/article/LeadingImage.js
+++ b/src/components/article/LeadingImage.js
@@ -1,16 +1,32 @@
+
 /* eslint no-unused-vars:0 */
 'use strict'
+import React from 'react' // eslint-disable-line
+import ReactDOM from 'react-dom'
 import classNames from 'classnames'
 import styles from './LeadingImage.scss'
 import { Image } from './Image'
-import React from 'react' // eslint-disable-line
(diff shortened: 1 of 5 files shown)


nickhsine:
@garfieldduck please help to review it

src/components/article/LeadingImage.js
```diff
@@ -34,12 +50,17 @@ class LeadingImage extends React.Component {
         break
     }
 
+    const defaultImgUrl = replaceStorageUrlPrefix(_.get(image, 'tiny.url'))
+
+    const imgJsx = this.state.isMount ? (
+      <img src={defaultImgUrl} alt={description} srcSet={getImageSrcSet(image)} />
+    ) : (
+      <img src={defaultImgUrl} alt={description} />
```
garfieldduck: do you want to add a blurry class/effect for the first loaded image?

nickhsine: The leading image will be a tiny image at first.

garfieldduck: Yes. I mean maybe you could add a blurry image class to the first loaded tiny image.

PR #586, opened 2017-08-04
Update Article and restore Footer on index

YuCJ:
- Untrack `.DS_Store` in static
- Restore Footer component on index
- Fix topic data in Article
- Update topicBlock of Article
- Add Mobile Article Tools
- Add Desktop Article Tools


src/containers/Article.js
```diff
@@ -53,12 +57,19 @@ const _ = {
...
 const { fetchAFullPost } = actions
 
+/* Issue need to be solved: _setArticleBounding doesn't get right endY if there are lazyload imgs or embedded items */
 let articlePostition = {
   beginY: 120,
   endY: 200,
   percent: 0
 }
 
+let scrollPosition = {
```
nickhsine: `const`

src/containers/Article.js
```diff
@@ -117,8 +128,9 @@ class Article extends Component {
     super(props, context)
 
     this._setArticleBounding = this._setArticleBounding.bind(this)
-    this._onScroll = this._onScroll.bind(this)
+    this._onScroll = _.throttle(this._onScroll, 200).bind(this)
```
nickhsine: Since you use throttle to decrease scroll events, you can consider remove the `raf`.
I think `raf` will be redundant in the case.

src/containers/Article.js
```diff
@@ -246,6 +260,11 @@ class Article extends Component {
     return top
   }
 
+  _onResize() {
+    this._setArticleBounding()
+    viewportHeight = window.innerHeight
```
nickhsine: Since you already listen on `resize` event, you can know the viewport width to decide to display the tools.
Hence, you can skip the calculation we don't need. For example, you can skip calculation for mobile device when the user is using desktop.

src/containers/Article.js
```diff
@@ -174,8 +187,9 @@ class Article extends Component {
   }
 
   componentWillUnmount() {
-    window.removeEventListener('resize', this._setArticleBounding)
+    window.removeEventListener('resize', this._onResize)
```
nickhsine: Since you add `throttle` here, could you kindly add throttle on resize event listener?

PR #591, opened 2017-08-08
Update Article components

YuCJ:
- Update article tool styles
- Remove lazyload on photograph articles
- Update ref to callback function
- Remove raf in Article
- Check screen type in Article
- Replace pure-render-mixin with PureComponent


nickhsine:
👍 💯

PR #613, opened 2017-08-15
[Update] registration system

hanyulo:


src/containers/SignUp.js
```diff
@@ -0,0 +1,20 @@
...
+    <SignUpForm
+      title={TITLE}
+      signUpMessage={SIGNUP_MESSAGE}
+      {...props}
+    />
+  </div>
+)
+
+
+export default connect()(SignUp)
```
nickhsine: use `styled-components`

src/containers/SignIn.js
```diff
@@ -0,0 +1,28 @@
...
+  <div className={styles['form-container']}>
+    <SignInForm
+      title={TITLE}
+      historyManager={browserHistory}
+      signInRedirectPath = {REDIRECT_PATH}
+      account={true}
+      google={true}
+      facebook={true}
+      defaultStyle={true}
+      {...props}
```
nickhsine: refine these props. delete those you don't need

server/server.js
```diff
@@ -61,6 +64,28 @@ server.get('*', async function (req, res, next) {
...
+    apiUrl: 'http://localhost:8080',
+    signUp: '/v1/signup',
+    signIn: '/v1/login',
+    activate: '/v1/activate',
+    oAuthProviders: {
+      google: '/v1/auth/google',
+      facebook: '/v1/auth/facebook'
+    },
+    location: 'http://testtest.twreporter.org:3000',
+    domain: 'twreporter.org'
```
nickhsine: define these fields into config

src/index.js
```diff
@@ -31,6 +32,21 @@ const store = configureStore(reduxState)
...
+const { auth } = store.getState()
+if(auth.authenticated && auth.authInfo && (auth.authType=== 'facebook' || auth.authType==='google')) {
+  setupTokenInLocalStorage(auth.authInfo)
+  store.dispatch(deletAuthInfoAction())
+}
+
+// 7 = 7 days
+store.dispatch(authUserByTokenAction(7, auth.authType))
+  .then(() => {})
+  .catch(() => {})
```
nickhsine: is this not complete yet?

src/index.js
```diff
@@ -31,6 +32,21 @@ const store = configureStore(reduxState)
...
 
+// Check if token existed in localStorage and expired
+// token can be stored in localStorage in two scenario
+// 1. oAuth
+// 2. TWReporter account sign in
+const { auth } = store.getState()
+if(auth.authenticated && auth.authInfo && (auth.authType=== 'facebook' || auth.authType==='google')) {
+  setupTokenInLocalStorage(auth.authInfo)
+  store.dispatch(deletAuthInfoAction())
+}
```
nickhsine: the codes and comments seems not to be consistent.

Is the business logic to set token into localStorage and delete it from redux state?
If so, why?

src/reducers/index.js
```diff
@@ -10,6 +11,23 @@ import twreporterRedux from 'twreporter-redux'
...
+  activate: '',
+  bookmarkUpdate: '',
+  bookmarkDelete: '',
+  bookmarkGet: '',
+  ping: '',
+  oAuthProviders: {
+    google: '',
+    facebook: ''
+  }
+}
```
nickhsine: refine this initial state

src/index.js
hanyulo: Sorry for the confusion. I update the comment and please check the comment again. 

Yes, The auth info pass from API to front end app by three mediums. Cookie -> Redux State -> localStorage. So, after the info package is stored in localStorage, I suppose that it should be removed from the redux state. BTW, it is only for oAuth .

src/index.js
hanyulo: I need the promise handler to prevent errors. But, I think that the front end app no need to do anything when the auth info is not existed or is expired in localStorage.

nickhsine: Since you don't do any business logic after the action, why do you need to make it as a `Promise`?

PR #628, opened 2017-08-21
Update Topics

YuCJ:
- Remove no-used initState in header reducer
- Update twreporter-redux
- Compose data in mapStateToProps instead
- Fix requests to fetch data


nickhsine:
🔥 :1st_place_medal:

PR #639, opened 2017-08-28
Update twreporter-react-footer-components

YuCJ:


YuCJ:
Change package from `twreporter-react-footer-components` to `twreporter-react-components`

PR #641, opened 2017-08-30
[Bug] Slideshow will switch to the correct one on Android

garfieldduck:
Slideshow will switch to the correct one on switching button clicked on Android

@nickhsine

### DIFF
--- a/src/components/article/slideshow/slideshow.js
+++ b/src/components/article/slideshow/slideshow.js
@@ -23,6 +23,7 @@ class Slideshow extends Component {
       currentIndex: 0,
       slideshowWidth: 0
     }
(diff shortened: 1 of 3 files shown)


src/components/article/slideshow/slideshow.js
```diff
@@ -21,7 +21,8 @@ class Slideshow extends Component {
     super(props, context)
     this.state = {
       currentIndex: 0,
-      slideshowWidth: 0
+      slideshowWidth: 0,
+      isSliding: false   // users are not able to switch slides if isSliding is true
```
nickhsine: do we need `isSliding` in state?
or just use global variable?

PR #673, opened 2017-10-02
Several updates. use npm-scope, fix bug, and add theme to every pages we owned.

nickhsine:
1. Use npm-scope and update dependenecies

2. Bug fix. Follow the docs to render React Router

3. Add theme for listing, article and photography, ...etc pages
  * Remove redux way to pass the theme instead of passing the theme as
component props
  * Use HOC, with-layout.js, to wrap the layout on any components

@garfieldduck @hanyulo @YuCJ


garfieldduck:
LGTM :shipit:

hanyulo:
👍

PR #699, opened 2017-12-01
Code refactors

nickhsine:
#  Do server side rendering by naive way.
Enable server side rendering by writing/reading the assets built by
webpack into files(./webpack-assets.json), and transpile server side es6 files, 
convert scss files into css modules and svg into data uri by babel plugin.

### DEVELOPMENT MODE
```
npm run dev
``` 
The above command will do the following two steps
1. Start `webpack-dev-server` to build/update/sync wepback bundles with clients.
2. Use `babel-node` to start dev server which **runtime** transpiles es6 files and converts scss, svg files by `babel-plugin-css-modules-transform` and `bable-plugin-inline-import-data-uri`.

### PRODUCTION MODE
```
npm run start
// which is equal to
// npm run build && npm run start-server
```
1. `npm run build` will 

- build webpack bundles 
- transpiles/converts es6, scss, svg files into `dist` folder which used on the server side.
2. Write webpack bundles file paths into `webpack-assets.json` which is required by `dist/server.js` so that server can populate the right assets path in the html.
3. Start the server by node, rather than babel-node or other runtime transpiling programs.

## CHANGES
- Remove webpack-isormorphic-tools
- Remove bootstrap and inject global styles by `styled-components`
- Run transipled js files on production, discard babel-node
- Upgrade webpack to ^3.0.0 and corresponding dependencies
- Replace `piping` by `nodemon` 

@YuCJ @hanyulo


YuCJ:
LGTM :+1:
We can test if there is any css bug on staging

webpack.config.js
```diff
@@ -0,0 +1,209 @@
...
+        assets.stylesheets.push(webpackConfig.output.publicPath + filename)
+        fs.writeFileSync('./webpack-assets.json', JSON.stringify(assets))
+        return filename
+      },
+      disable: false,
+      allChunks: true
+    }),
+    new webpack.optimize.UglifyJsPlugin({
+      compress: {
+        warnings: false
```
YuCJ: the default value of `compress.warning` is `false` in webpack@^2
https://webpack.js.org/guides/migrating/#uglifyjsplugin-warnings

YuCJ:
`npm run dev` with `concurrently@^0.1.1` throws an error. It works with `concurrently@3.5.1`

PR #720, opened 2017-12-18
Turn off css source map in production and bump twreporter-react-component version to 2.0.6

YuCJ:
- Turn off css source map with webpack in production
- Bump `twreporter-react-component` version to 2.0.6


nickhsine:
LGTM

PR #813, opened 2018-02-06
Add react-loadable and decrease BlockQuote font-size

YuCJ:
#### Patches
- Decrease BlockQuote font-size to 20px
- Add dynamic import with react-loadable


nickhsine:
We need to make sure those components used frequencies.
May not make all of them loadable.

nickhsine:
LGTM 👍  BRAVO

PR #822, opened 2018-03-08
Update @twreporter/react-components package version

taylrj:
Update @twreporter/react-components package version from 4.0.1 to 4.0.3.


nickhsine:
LGTM

PR #867, opened 2018-04-16
Adopt `localforage` which uses async storage(indexedDB or WebSQL) and fallback with localStorage

nickhsine:
@YuCJ @taylrj 

PWA documents suggest that not  to use `localStorage` on production because `localStorage` API is synchronous.

Therefore, before adopting PWA, replace `localStorage` by `localforage` which uses `indexedDB` by default,  and fallback to `localStorage`.


YuCJ:
LGTM 👌

taylrj:
LGTM

PR #876, opened 2018-05-02
[PWA] - Adopt Service Worker 

nickhsine:
## PWA Features 
- [x] Progressive - render App Shell, like header and footer, first. And then render the content of the page.
- [x] Responsive - Fits any form factor: desktop, mobile, tablet, or whatever is next.
- [x] Connectivity independent - work offline or on low-quality networks
- [x] Discoverable - Is identifiable as an "application" thanks to W3C manifest and service worker registration scope, allowing search engines to find it.
- [x] Installable - Allows users to add apps they find most useful to their home screen without the hassle of an app store.
- [x] Fresh - Always up-to-date thanks to the service worker update process.
- [x] Safe - Served via HTTPS to prevent snooping and to ensure content hasn't been tampered with.
- [ ] App-like - Feels like an app, because the app shell model separates the application functionality from application content .
- [ ] Re-engageable - Makes re-engagement easy through features like push notifications.
- [ ] Linkable -  Easily share the application via URL, does not require complex installation.

### What does this PR do ?
The above checked options should be implemented by this PR.

In this PR, `generate-service-worker.js` use `sw-precache` to generate the service worker file, named `/sw.js`,  served by express server.

When it generates service worker file, it will not only use `webpack-assets.json` to define what static files should be cached, but also define the dynamic URL patterns which will be applied on `fetch` event, tell the browser to fetch and cache the dynamic content.

Service worker provides the fallback rule, using the client side rendering static content served at `/sw-fallback-page`, for the browser when the resource can not be loaded yet.


taylrj:
LGTM

PR #878, opened 2018-05-08
 [Service Worker] Remove `sw-precache`. Build the service worker by template file.

nickhsine:
## What does this PR do?
The [previous PR](https://github.com/twreporter/twreporter-react/pull/877) works abnormally (wrong layout) when offline. Hence, I replace [sw-precache](https://github.com/GoogleChromeLabs/sw-precache/)(a script to generate service worker) by our own script.

## How does this PR do?
This PR learns the way `sw-precache`  generates service worker,  using the template file(service-worker/service-worker.tmpl) and `sw-toolbox` to generate custom service worker file.

## Increased endpoints of this PR
1) Generate service worker file by `make build` | `make start`(if you want to start the server to test). The generated file, named `sw.js`, is hosted on `/sw.js` endpoint by express server.

2) Add `/sw-fallback-page` endpoint into the react routers. This endpoint is used to serve(network failure notification) users when networks disconnected.

## How does sw.js treat the request and response?
1) Service worker will cache static files, such as webpack bundles, when the users connect to our site. The cache key is a hash value according to `webpack-asset.json`. Therefore, if you rebuild the webpack bundles,  the cache key will change. The users won't use the old cached webpack bundles to render the page.

2) In the browser `fetch` event listener, we use [networkFirst](https://jakearchibald.com/2014/offline-cookbook/#network-falling-back-to-cache) policy. The request always go to the network, save the response into cache if needed,  and use cached entry if request fails.

3) Except for static files, the following URL patterns(request) we will cache the response and use the cache to serve
- `/^\/(a|topics|tags|category|categories)\//` (article, topic, listing page)
- `/^\/$/` (index page)
- `/^\/photography/` (photography page)
- `/v1/posts` (api endpoint for posts)
- `/v1/topics` (api endpoint for topics)
- `/v1/index_page(_categories)?` (api endpoints for index page)

## The Difficulties
Since our web site is a SPA(single page application), we don't have any static files, such as `index.html`, to serve. Hence, when the user links to `/a/test-slug` from `/categories/reviews`, the browser won't send the request(`/a/test-slug`) to the server, on the other hand, the browser only send the api request to fetch the data. In this situation, we could not cache the response of `/a/test-slug`.

If we could not cache the `/a/test-slug`, how can we serve the content when users is offline?

We serve an new endpoint, `/sw-fallback-page`, as a fallback page, which is nothing but a client side rendering page. It will notice the users the  network disconnect,  and let users to click the link, like `/a/test-slug`, to render the page on client side. Since we will cache the response from api server, so the client side rendering will work fine as we expect.

@YuCJ @taylrj


taylrj:
LGTM 👌

PR #879, opened 2018-05-09
Adds some space between channel bar and title in article page on tablet

taylrj:
This patch adds margin space between channel bar and title on tablet. The new added margin-top value equals to the margin-bottom value of title.


nickhsine:
LGTM 👍

PR #886, opened 2018-05-20
Bump to v3.0.0. Adopt Service Worker.

nickhsine:
## PWA Features 
- [x] Connectivity independent - work offline or on low-quality networks
- [x] Installable - Allows users to add apps they find most useful to their home screen without the hassle of an app store.
- [x] Re-engageable - Makes re-engagement easy through features like push notifications.

### What does this PR do ?
The above checked options should be implemented by this PR.

### Service Worker
This PR learns the way `sw-precache`  generates service worker,  using the template file(service-worker/service-worker.tmpl) and `sw-toolbox` to generate custom service worker file.

When it generates service worker file, it will not only use `webpack-assets.json` to define what static files should be cached, but also define the dynamic URL patterns which will be applied on `fetch` event, tell the browser to fetch and cache the dynamic content.

Service worker provides the fallback rule, using the client side rendering static content served at `/sw-fallback-page`, for the browser when the resource can not be loaded yet.

### New endpoints of this PR
1) Generate service worker file by `make build` | `make start`(if you want to start the server to test). The generated file, named `sw.js`, is hosted on `/sw.js` endpoint by express server.

2) Add `/sw-fallback-page` endpoint into the react routers. This endpoint is used to serve(network failure notification) users when networks disconnected.

### How does sw.js treat the request and response?
1) Service worker will cache static files, such as webpack bundles, when the users connect to our site. The cache key is a hash value according to `webpack-asset.json`. Therefore, if you rebuild the webpack bundles,  the cache key will change. The users won't use the old cached webpack bundles to render the page.

2) In the browser `fetch` event listener, we use [networkFirst](https://jakearchibald.com/2014/offline-cookbook/#network-falling-back-to-cache) policy. The request always go to the network, save the response into cache if needed,  and use cached entry if request fails.

3) Except for static files, the following URL patterns(request) we will cache the response and use the cache to serve
- `/^\/(a|topics|tags|category|categories)\//` (article, topic, listing page)
- `/^\/$/` (index page)
- `/^\/photography/` (photography page)
- `/v1/posts` (api endpoint for posts)
- `/v1/topics` (api endpoint for topics)
- `/v1/index_page(_categories)?` (api endpoints for index page)

### The Difficulties
Since our web site is a SPA(single page application), we don't have any static files, such as `index.html`, to serve. Hence, when the user links to `/a/test-slug` from `/categories/reviews`, the browser won't send the request(`/a/test-slug`) to the server, on the other hand, the browser only send the api request to fetch the data. In this situation, we could not cache the response of `/a/test-slug`.

If we could not cache the `/a/test-slug`, how can we serve the content when users is offline?

We serve an new endpoint, `/sw-fallback-page`, as a fallback page, which is nothing but a client side rendering page. It will notice the users the  network disconnect,  and let users to click the link, like `/a/test-slug`, to render the page on client side. Since we will cache the response from api server, so the client side rendering will work fine as we expect.

### Installable - Add manifest.json
Provide manifest and `<meta>` tag to tell the mobile browser to add the application to the home screen. 
More detail can be found in [the doc](https://developers.google.com/web/fundamentals/app-install-banners/)

### Web Push Subscription and Notification
In `src/container/App.js`, we render a box to notify users that we have Notification service right now. If users click accept, and then the service worker will send requests to backend to subscribe web push notifications. 
Service worker will listen to the `push` event and `notificationclick` event. Therefore, we can push notifications to the browsers from remote servers.

#### TODO List
1)  Beginning in Chrome 68 (May 2018), Chrome will not automatically show the Add to Home Screen prompt, instead, we must trigger it by using the `beforeinstallprompt` event.
2) Need to clean up the web push subscription periodically since the subscriptions may expire.

@YuCJ @taylrj @babygoat


src/client.js
```diff
@@ -24,6 +31,113 @@ function scrollToTopAndFirePageview() {
...
+  for (let i = 0; i < rawData.length; ++i) {
+    outputArray[i] = rawData.charCodeAt(i)
+  }
+  return outputArray
+}
+
+/**
+ * Check if the web push endpoint is subscribed o not
+ * @param {string} endpoint Web push subscription endpoint
+ * @param {Promise} A Promise resolves to true if subscribed, otherwise, false.
```
taylrj: It should be `@return {Promise}` ?

taylrj:
LGTM 👍 👍 👍

src/containers/ServiceWorkerFallbackPage.js
```diff
@@ -0,0 +1,62 @@
...
+`
+
+class FallbackPage extends Component {
+  //componentDidMount() {
+  //  const pathname = _.get(this.props, 'location.pathname')
+  //  console.error('pathname:', pathname)
+  //  if (pathname !== '/sw-fallback-page') {
+  //    this.props.push(`/sw-fallback-page?redirect=${pathname}`)
+  //  }
+  //}
```
YuCJ: Is there any situation we need to push new location? or can this be removed?

nickhsine: it can be removed.

nickhsine:
@YuCJ @taylrj Please help to review the latest commit.

taylrj:
LGTM 👍

src/server.js
```diff
@@ -193,6 +193,7 @@ pe.skipPackage('express')
 
 app.use((err, req, res, next) => { // eslint-disable-line no-unused-vars
   console.log(pe.render(err)) // eslint-disable-line no-console
+  res.header('Cache-Control', 'not-store')
```
YuCJ: typo of `no-store`?

nickhsine: yes...a huge typo

PR #916, opened 2018-06-20
Bug fix and service worker boilerplate improvement

nickhsine:
### Bug Description
Since we adopted service worker, `embed.js` of `venngage` works abnormally.
It won't generate the iframe even if we actually render the 
```
// script attributes are given by venngage
<script 
async="true"
src="https://infograph.venngage.com/js/embed/v1/embed.js" 
data-vg-id="KTf59BxqBAA" 
data-title="sexual_assault"
data-w="700"
data-h="525" 
data-multipage="true"
data-f="false">
</script>
```

### Details 
Because service worker caches the import resources, such as stylesheets and js bundles, window dispatch `load` event early than the time venngage script downloaded and executed. 
And in venngage script(https://infograph.venngage.com/js/embed/v1/embed.js), the script generates the `iframe` by listening to `load` event of `window`.
Therefore,  venngage script won't generate the `iframe`.

### Solution
We use a workaround to solve this problem.
The solution is adding a callback function, which dispatches `load` event,  after venngage script loaded.

@taylrj @YuCJ

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -49,16 +49,66 @@ configureStore(browserHistory, reduxState)
     })
   })
 
-// Setup service worker if browser supports
-if('serviceWorker' in navigator && !__DEVELOPMENT__) {
-  // register service worker
-  navigator.serviceWorker.register('/sw.js').then(function (registration) {
-    // Registration was successful
-    console.log('ServiceWorker registration successful with scope: ', registration.scope)
-  }, function (err) {
-    // registration failed :(
-    console.log('ServiceWorker registration failed: ', err)
+
+/**
+ * Copyright 2015 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


src/client.js
```diff
@@ -49,16 +49,66 @@ configureStore(browserHistory, reduxState)
...
+      // updatefound is fired if service-worker.js changes.
+      reg.onupdatefound = function () {
+        // The updatefound event implies that reg.installing is set; see
+        // https://w3c.github.io/ServiceWorker/#service-worker-registration-updatefound-event
+        const installingWorker = reg.installing
+
+        installingWorker.onstatechange = function () {
+          switch (installingWorker.state) {
+            case 'installed':
+              if (navigator.serviceWorker.controller) {
```
YuCJ: I've found a list about the lifetime of service-workers, but didn't get why that `navigator.serviceWorker.controller` exists means there is new content
https://developers.google.com/web/fundamentals/primers/service-workers/lifecycle#_8

PR #925, opened 2018-06-29
Fix circleci build failing. Install the corresponding kubectl pkg

nickhsine:
The latest version of kubectl(v1.11.0) breaks the circle builds.
Before our containers upgrades to v1.11.0, we remain kubectl clients on v1.8.7.

@babygoat

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,43 +55,52 @@ jobs:
           keys:
             - v1-build-{{ checksum ".circle-sha" }}
       - run:
-          name: Install kubectl
+          name: Connect to container cluster
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -55,43 +55,52 @@ jobs:
...
             fi
 
-
             echo "CLUSTER_NAME: ${CLUSTER_NAME}"
+            gcloud --quiet container clusters get-credentials $CLUSTER_NAME
+
+      - run:
+          name: Install kubectl
+          command: |
+            KUBECTL_VERSION="$(gcloud container clusters describe twreporter-website | sed -n 's/.*version:[ ]*\(.*\)-\(.*\)/\1/p')"
```
babygoat: use $CLUSTER_NAME instead of hard coding the cluster name?

nickhsine: ok. will change it.

PR #928, opened 2018-07-11
fix(ui): quick fix for desktop tool position

ArvinH:
Sorry, just can’t bear seeing this lol
However, this is a quick fix, it must have a better way to fix this.

before:
image

after:
image


nickhsine:
Thanks for reporting and sending PR.
We will fix this in next release.

PR #947, opened 2018-07-19
Fix babel plugin order causing ssr error and modify .babelrc to make it more readible

YuCJ:
Reference:
official document says:
>  The plugin call order in your .babelrc file matters. If you're using the env property in your babel configuration, then putting this plugin into the plugins array won't suffice. Instead it needs to be put into each env's plugins array to maintain it being executed first.
https://www.styled-components.com/docs/tooling#better-debugging

also can see:
https://github.com/styled-components/babel-plugin-styled-components/issues/78#issuecomment-361160935


nickhsine:
We may use (creating-a-preset)[https://babeljs.io/docs/en/plugins/#creating-a-preset] to create runtime configurable plugins.

nickhsine:
LGTM 👍

PR #948, opened 2018-07-20
Fix selection styles and make several fixes on topic landing page

YuCJ:
Topic landing page:
- Move Description out of Topic module and fix styles bug
- Refator Topic and the backgrounds and banner of it
- Fix the problem that Banner height not fits with viewport height


nickhsine:
LGTM 👍

PR #950, opened 2018-07-23
[Bug] Prevent window from scrolling when an overlay modal opens up on about-us page

taylrj:
This patch includes following changes :
- Prevent window from scrolling when an overlay modal opens up on about-us page
- Add a new member
- Setting the height of opening section to auto instead of `height: 100vh`


src/components/about-us/utils/pop-up-panel.js
```diff
@@ -0,0 +1,116 @@
...
+
+  _handlePreventTouchendWhenPanning(event) {
+    this.panel.scrollTop = this.panel.scrollTop + (this.startY - event.changedTouches[0].screenY)
+  }
+
+  _handlePreventTouchmoveWhenPanning = (event) => {
+    event.preventDefault()
+    this.panel.scrollTop = this.panel.scrollTop + (this.startY - event.changedTouches[0].screenY)
+    this.startY = event.touches[0].screenY
+  }
```
nickhsine: We may need to figure out the difference between `touches` and `changedTouches`.

taylrj: In our case, `touches` and `changedTouches` are exactly the same when `touchmove` event is triggered.

PR #953, opened 2018-07-24
 [Bug]Adjust window height of pop-up-panel dynamically on about-us page

taylrj:
This patch adds an event listener for the `onResize` event to change window height when the URL bar is collapsed. This situation was found in chrome browser of android devices.


nickhsine:
LGTM

PR #960, opened 2018-07-26
Update React-Loadable and EmbeddedCode

YuCJ:
Fix venngage embedded code was replace when embedded code is updated:
- Save return value of Loadable into closure to prevent creating new Component every time `getComponent` is called
- Update EmbeddedCode to PureComponent to prevent rerender when props not changed


nickhsine:
LGTM 💯

PR #971, opened 2018-07-31
[Bug] Prevent from changing original data object in section2 on about-us page

taylrj:
[Update] Use Object.assign to create a new object in order not to modify the original data `groupedMembers`


nickhsine:
LGTM 👍

PR #974, opened 2018-07-31
Release 3.0.14 on production

taylrj:


YuCJ:
@taylrj I've confirmed that the changes in files except `about-us` are all in v3.0.12

PR #983, opened 2018-08-21
Minor changes of footer after testing

taylrj:
This patch aims to use full absolute path instead of relative path in static assets. This commit moves static assets of footer to a specific folder in gcs. Therefore, the static assets can be linked by a url link, and one can get these resources whether under `www.twreporter.org` or not.


taylrj:
@nickhsine

PR #986, opened 2018-08-22
[Bug] Fix background color of footer in article pages and remove static file path from props of Footer component

taylrj:
This patch adds static files prefix to pages footer apart from homepage. And uses default theme color as the constant `props` `bgColor`which is passed to Footer component.


taylrj:
[This patch](https://github.com/twreporter/twreporter-react-components/pull/112) fixes the comment above

PR #1050, opened 2018-10-12
Add GA events to donation button in article pages

taylrj:
This PR adds ga click event to button in article donation box and is different from [that PR](https://github.com/twreporter/twreporter-react-components/pull/118), which adds click events to donation section in homepage and footer.


nickhsine:
LGTM

PR #1083, opened 2019-01-11
Tweak TopicLandingPage container due to api changes

taylrj:
Since go-api will remove html transformation field in next version, the topic landing page on main site need to be updated along.
This patch contains topic landing page updates listed in the following bullets : 
- `description` and `teamDescription` in TopicLandingPage container will receive api_data as an array instead of plain html
- Reuse `Body` component which has been used in `Article` to be the content layout of landing page


taylrj:
@nickhsine @babygoat

src/components/topic/Description.js
```diff
@@ -56,21 +46,24 @@ const TopicDescription = styled.div`
     margin: auto;
     border-top: 2px solid ${colors.red.rustyRed};
   }
+
+  /* Since the component corresponding to unstyled type uses a className to address text-align as justify. 
+   * Boost the specificity of the override style here */
+  && * {
+    font-size: ${typography.font.size.base};
+    line-height: ${lineHeight.linHeightLarge};
```
babygoat: Minor. I suggest correcting the variable name typo as well.(`linHeightLarge` -> `lineHeightLarge`)

src/components/topic/Description.js
```diff
@@ -56,21 +46,24 @@ const TopicDescription = styled.div`
     margin: auto;
     border-top: 2px solid ${colors.red.rustyRed};
   }
+
+  /* Since the component corresponding to unstyled type uses a className to address text-align as justify. 
+   * Boost the specificity of the override style here */
```
nickhsine: May we change commenting style to
```
/**
 *  Since the component corresponding to unstyled type uses a className to address text-align as justify. 
 *  Boost the specificity of the override style here
 */
```
or 
```
// Since the component corresponding to unstyled type uses a className to address text-align as justify. 
// Boost the specificity of the override style here
```

src/components/topic/Description.js
```diff
@@ -56,21 +46,24 @@ const TopicDescription = styled.div`
...
+  && * {
```
nickhsine: Do we need `&&`  here? or maybe we just need `&` ?

taylrj: After the trial, I think `&` is not enough to override the specificity of className `align-justify`

src/components/topic/Description.js
taylrj: Sure.

PR #1093, opened 2019-01-25
Fix many components in topic landing page after testing on staging

taylrj:
This patch aims to fix topic landing page which has been merged in this [PR](https://github.com/twreporter/twreporter-react/pull/1083) and was reverted after testing on staging server.

## Change List
- Removed the `Body` component which is used in article pages and reused `getArticleComponent` instead
- Fix  components returned in type `blockquote`, `youtube`, `image` in team description, and abnormal "double" underline hyper links in topic description texts
- Fix the typo `linHeightLarge` in `common-variables`

### DIFF
--- a/src/components/topic/Description.js
+++ b/src/components/topic/Description.js
@@ -1,37 +1,23 @@
+import PropTypes from 'prop-types'
 import React from 'react'
 import YoutubePlayer from 'react-youtube'
-import constStyledComponents from '../../constants/styled-components'
 import getArticleComponent from '../article/getArticleComponent'
 import styled from 'styled-components'
-import PropTypes from 'prop-types'
 import { lineHeight, typography, colors } from '../../themes/common-variables'
+import { screen } from '../../themes/screen' 
 // lodash
 import get from 'lodash/get'
 
-const fontSize = 'medium'
+const mediumFontSize = 'medium'
 const redLineDistance = '40px'
-const contentSize = 'small'
+const smallContentSize = 'small'
 
 const _ = {
   get
 }
 
-const StyledTopicComponent = constStyledComponents.ResponsiveContainerForAritclePage.extend`
-  font-size: ${(props) => {
-    let fontSize = typography.font.size.base
-    switch(props.fontSize) {
-      case 'small':
-        fontSize = typography.font.size.xSmall
-        break
-      case 'large':
-        fontSize = typography.font.size.larger
-        break
-      default:
-        break
-    }
-    return fontSize
-  }};
-
+const StyledTopicComponent = styled.div`
+  font-size: ${props => props.fontSize};   
   margin-bottom: 40px;
 `
 
@@ -44,6 +30,8 @@ const Container = styled.div`
 
 const TopicDescription = styled.div`
   position: relative;
+  width: 38rem;
+  max-width: 90%;
   padding-bottom: ${redLineDistance};
   text-align: center;
   margin-left: auto;
@@ -52,7 +40,6 @@ const TopicDescription = styled.div`
   white-space: pre-wrap;
   line-height: ${lineHeight.linHeightLarge};
   color: ${colors.gray.gray15};
-
   p {
     margin-bottom: 1.5em;
   }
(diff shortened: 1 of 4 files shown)


src/components/topic/Description.js
```diff
@@ -81,21 +114,168 @@ const TeamDescriptiton = styled.div`
...
+const TopicTeamJSX = {
+  Image: (content) => {
+    const imgSrc = _.get(content, [ 0, 'tablet', 'url' ], '')
+    return (
+      <img src={imgSrc} />
+    )
+  }
+}
+
+const getTopicContent = (data, contentType) => {
```
nickhsine: Do we need to know `contentType`?
How about handling `image`, `youtube` and `blockquote` by default?
So we can remove the condition
```
if(contentType === 'description') {
} else if(contentType === 'team'){
}
```

src/components/topic/Description.js
```diff
@@ -81,21 +114,168 @@ const TeamDescriptiton = styled.div`
...
+            TopicComponent = TopicTeamJSX.Image
+            break
+          default:
+            ArticleComponent = getArticleComponent(type)
+        }
+      }
+
+      if (!ArticleComponent && !TopicComponent) {
+        return null
+      }
```
nickhsine: Do we need  both `ArticleComponent` and `TopicComponent` here?
Could we only use `Component` instead?

src/components/topic/Description.js
```diff
@@ -81,21 +114,168 @@ const TeamDescriptiton = styled.div`
...
+          size={contentSize}
+        >
+          {
+            ArticleComponent ?
+              <ArticleComponent
+                alignment={ele.alignment}
+                content={ele.content}
+                id={ele.id}
+                styles={ele.styles}
+              />:TopicComponent(_.get(ele, 'content', []))
```
nickhsine: Follow the above comments,
could we rewrite it like
```
<Component
  alignment={ele.alignment}
  content={ele.content}
  id={ele.id}
  styles={ele.styles}
/>
```

src/components/topic/Description.js
```diff
@@ -1,22 +1,49 @@
...
+        fontSize = typography.font.size.xSmall
+        break
+      case 'large':
+        fontSize = typography.font.size.larger
+        break
+      default:
+        break
+    }
+    return fontSize
+  }};
```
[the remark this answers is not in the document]
taylrj: We may not need the switch case since the fontSize is fixed in this page, I will remove that in next update

src/components/topic/Description.js
```diff
@@ -1,22 +1,49 @@
...
 
+const fontSize = 'medium'
 const redLineDistance = '40px'
+const contentSize = 'small'
+
+const _ = {
+  get
+}
+
+const StyledTopicComponent = constStyledComponents.ResponsiveContainerForAritclePage.extend`
```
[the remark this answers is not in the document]
taylrj: Will remove style in `constStyledComponents.ResponsiveContainerForAritclePage` from `StyledTopicComponent` since layout width of topic landing page is inconsistent with our article page in tablet. I had not noticed that, it's my fault.

src/components/topic/Description.js
```diff
@@ -81,21 +114,168 @@ const TeamDescriptiton = styled.div`
...
+    margin-bottom: 0;
+  }
 `
 
+const YoutubeIframeContainer = styled.div`
+  max-width: 100%;
+  margin-bottom: 0;
+  iframe,object,embed {
+    width: 560px;
+    height: 315px;
```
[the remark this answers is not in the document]
taylrj: The parent of `iframe` has style `max-width: 100%` and the width in mobile is no more than 100% width. However, code `width: 560px` here seems like a bug and not clear enough to read. I will rewrite the `iframe` style to specify `560px` width only to devices width larger than and equals to tablets. Thanks for the reminder.

src/components/topic/Description.js
taylrj: Sure. Will follow these suggestions in next update.

taylrj:
updated

nickhsine:
LGTM

PR #1102, opened 2019-02-13
Bugfix and update in about us page

taylrj:
## Change Log
- This patch fixes about-us page opening layout which got wrong height (100 viewport height) when client width is smaller than desktop and larger than tablet breakpoint.
- Update member information (adds a new member 劉致昕) in section2


nickhsine:
LGTM.
Please also modify `CHANGELOG.md` and `packge.json` if needed.

PR #1108, opened 2019-02-20
Server side rendering and code refactoring of Bookmark

YuCJ:
- [x] bookmarks redux actions/reducers
- [x] bookmark list container
- [x] bookmark widget on article page container
- [x] bookmarks server side data loader
- [x] redirect user to login page when user is not authenticated on both client and server side rendering
- [ ] bookmarks redux actions/reducers unit tests


src/actions/error-action-creators.js
```diff
@@ -0,0 +1,80 @@
...
+    type: failActionType,
+    payload: {
+      message: err.message
+    }
+  }
+  if (err.response) {
+    if (err.response.status === statusCode.badRequest) {
+      return _.merge({}, failActionSharedProperties, {
+        message: 'response bad request due to data validation failures',
+        data: _.get(err, 'response.data')
```
nickhsine: miss `payload` here?

src/actions/error-action-creators.js
```diff
@@ -0,0 +1,80 @@
...
+  } else if (err.request) {
+    return _.merge({}, failActionSharedProperties, {
+      payload: {
+        message: 'request was made but no response was received'
+      }
+    })
+  } else {
+    return _.merge({}, failActionSharedProperties, {
+      payload: {
+        message: 'error happened in setting up the request that triggered an error'
```
nickhsine: Do we need merge here?
Since` failActionSharedProperties` is a pretty simple object,
could we just write the return object directly, like 
```
return {
  type: failActionType,
  payload: {
    data: _.get(err, 'response.data'),
    message: _.get(err, 'response.data.message'),
    statusCode: err.response.status
  }
}
```

src/constants/action-types.js
```diff
@@ -54,3 +54,17 @@ export const SET_PROGRESS_PRECENTAGE = 'SET_PROGRESS_PRECENTAGE'
...
+export const FETCH_MULTIPLE_BOOKMARKS_FAILURE = 'FETCH_MULTIPLE_BOOKMARKS_FAILURE'
+export const FETCH_SINGLE_BOOKMARK_REQUEST = 'FETCH_SINGLE_BOOKMARK_REQUEST'
+export const FETCH_SINGLE_BOOKMARK_SUCCESS = 'FETCH_SINGLE_BOOKMARK_SUCCESS'
+export const FETCH_SINGLE_BOOKMARK_FAILURE = 'FETCH_SINGLE_BOOKMARK_FAILURE'
+export const DELETE_SINGLE_BOOKMARK_REQUEST = 'DELETE_SINGLE_BOOKMARK_REQUEST'
+export const DELETE_SINGLE_BOOKMARK_SUCCESS = 'DELETE_SINGLE_BOOKMARK_SUCCESS'
+export const DELETE_SINGLE_BOOKMARK_FAILURE = 'DELETE_SINGLE_BOOKMARK_FAILURE'
+export const CREATE_SINGLE_BOOKMARK_REQUEST = 'CREATE_SINGLE_BOOKMARK_REQUEST'
+export const CREATE_SINGLE_BOOKMARK_SUCCESS = 'CREATE_SINGLE_BOOKMARK_SUCCESS'
+export const CREATE_SINGLE_BOOKMARK_FAILURE = 'CREATE_SINGLE_BOOKMARK_FAILURE'
```
nickhsine: Could we use camelCase here?
Although the old action type conventions are ALL UPPER CASES,
we could have a new convention since this PR.

YuCJ: I try to replace the ALL_UPPER_CASES style to camelCase. But I found that ALL_UPPER_CASES has better readability for action types because the request, failure, and success of a given async action looks so similar.

```js
// actions on bookmarks
export const fetchMultipleBookmarksRequest = 'FETCH_MULTIPLE_BOOKMARKS_REQUEST'
export const fetchMultipleBookmarksSuccess = 'FETCH_MULTIPLE_BOOKMARKS_SUCCESS'
export const fetchMultipleBookmarksFAILURE = 'FETCH_MULTIPLE_BOOKMARKS_FAILURE'
export const fetchSingleBookmarkRequest = 'FETCH_SINGLE_BOOKMARK_REQUEST'
export const fetchSingleBookmarkSuccess = 'FETCH_SINGLE_BOOKMARK_SUCCESS'
export const fetchSingleBookmarkFailure = 'FETCH_SINGLE_BOOKMARK_FAILURE'
export const deleteSingleBookmarkRequest = 'DELETE_SINGLE_BOOKMARK_REQUEST'
export const deleteSingleBookmarkSuccess = 'DELETE_SINGLE_BOOKMARK_SUCCESS'
export const deleteSingleBookmarkFailure = 'DELETE_SINGLE_BOOKMARK_FAILURE'
export const createSingleBookmarkRequest = 'CREATE_SINGLE_BOOKMARK_REQUEST'
export const createSingleBookmarkSuccess = 'CREATE_SINGLE_BOOKMARK_SUCCESS'
export const createSingleBookmarkFailure = 'CREATE_SINGLE_BOOKMARK_FAILURE'
```

src/actions/error-action-creators.js
YuCJ: I'll replace it to literal declaration.

src/constants/action-types.js
nickhsine: If we use nested objects to represent them?
such  as 
```
// constants/action-types.js
export const singleBookmark = {
    read: {
      request: 'request to fetch single bookmark',
      success: 'succes to fetch single bookmark',
      failure: 'fail to fetch single bookmark'
    },
   create: {
      request: 'request to  create single bookmark',
      success: 'succes to create single bookmark',
      failure: 'fail to create single bookmark'
   },
   delete: {
      request: 'request to  delete single bookmark',
      success: 'succes to delete single bookmark',
      failure: 'fail to delete single bookmark'
   }
}

// actions/bookmarks.js
import { singleBookmark as singleBookmarkAT }  from '../constants/action-types'
store. dispatch({
  actionType: singleBookmarkAT.read.request,
})

```

How do you think?
Will it be easier to understand?

src/actions/bookmarks.js
```diff
@@ -0,0 +1,254 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${jwt}`
+      }
+    }
+    return httpClient.delete(url, axiosConfig)
+      .then((res) => {
+        const successAction = buildSuccessActionFromRes(res, types.DELETE_SINGLE_BOOKMARK_SUCCESS)
+        successAction.payload.bookmarkID = bookmarkID
+        successAction.payload.byWidget = byWidget
```
nickhsine: Would the reducer handle `payload` in different ways due to `byWidget`?
Not very sure why we need `byWidget` in our payload.

src/constants/action-types.js
YuCJ: Nice

YuCJ:
Update bookmark list and bookmark widget containers

src/actions/bookmarks.js
YuCJ: legacy code forgot to remove

src/actions/error-action-creators.js
```diff
@@ -63,13 +62,15 @@ function handleAxiosError(err = {}, failActionType) {
       }
     }
   } else if (err.request) {
+    console.log('An axios error occured:', err) // eslint-disable-line
```
nickhsine: `axios.err` could be a very big object.
Printing it here seems not to be helpful for debugging.
How about adding `err.config` in the return object?
Let the caller decide if to print the log or not?
How do you think?

YuCJ: It looks fine. I'll update the error action payload as axios document. https://github.com/axios/axios#handling-errors

YuCJ:
All updated

src/client.js
```diff
@@ -2,14 +2,17 @@
...
+import configureStore from './store/configure-store'
 import Loadable from 'react-loadable'
 import React from 'react'
 import ReactDOM from 'react-dom'
 import ReactGA from 'react-ga'
-import configureStore from './store/configure-store'
-import { BrowserRouter, Route } from 'react-router-dom'
-import { Provider } from 'react-redux'
+
+const releaseBranch = config.releaseBranch
```
nickhsine: For client side, we might need to create `src/global-env.js`[(example)](https://github.com/twreporter/membership-fe/blob/master/src/global-env.js), since `config.js` contains cookie secret which is supposed not to be exposed on the client side.

src/actions/bookmarks.js
```diff
@@ -0,0 +1,294 @@
...
+  }
+}
+
+/**
+ *  @typedef {Object} FailAction
+ *  @property {string} type - Action error type
+ *  @property {string} payload
+ *  @property {string} payload.statusCode - Response status code
+ *  @property {string} payload.message - Error message
+ *  @property {object} payload.error - Error object
```
nickhsine: Remove `FailAction`? 
I didn't find where we use it.

src/actions/bookmarks.js
```diff
@@ -0,0 +1,294 @@
...
+ * @property {string} published_date
+*/
+
+/**
+ * @typedef {Object} RequestAction
+ * @property {string} type - The type of action
+ * @property {string} payload.method - The HTTP request method
+ * @property {string} payload.url - The target to send request
+ * @property {Object} payload.body - The request body with POST, PUT, DELETE, or PATCH request
+ * @property {Object} payload.headers - The request header
```
nickhsine: Remove `RequestAction`?

src/actions/bookmarks.js
```diff
@@ -0,0 +1,294 @@
...
+        payload: {
+          method,
+          url,
+          headers,
+          body: data
+        }
+      })
+      axiosInstanceWithUserCookie.interceptors.request.eject(interceptor)
+      return config
+    })
```
nickhsine: Interceptor is a good approach.
However, it is a one-time-used interceptor, which kind of loses its advantages.
Could we `dispatch(RequestAction)` directly?

src/client.js
YuCJ: Fixed

src/actions/bookmarks.js
YuCJ: Because we may set the `headers` via the global axios generator, the custom axios instance and the methods of the instance, use the interceptor is the most certain way to catch the `headers` that axios actually use to send a request (rather than merge the configs above manually)

PR #1138, opened 2019-03-18
Bugfix

YuCJ:
@nickhsine 
- Fix the bug of getSignInHref
- Fix topic header external link
- Fix topic arrow position


nickhsine:
LGTM
Please rebase those commits before merge.

PR #1164, opened 2019-05-16
Send Google Analytics pageview event by url instead of pathname

taylrj:
This patch revises ga pageview event to send by url instead of pathname so as to avoid the ambiguity between [support|account] pathnames and [www] pathnames

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## Unreleased
+- Send Google Analytics pageview event by url instead of pathname
+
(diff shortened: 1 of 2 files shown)


nickhsine:
Since the majority of ga requests are from main site, we could record the pathname only.

How do you think?

taylrj:
Since urls from our main site could be very long depends on its slug length, to keep it in short by recording pathname only is a better way for analysing.

This PR will be closed.

PR #1172, opened 2019-05-31
Set `embeddedCode` component to be printable by default

taylrj:
[Kanban Flow Card](https://kanbanflow.com/t/3Nk9smUx)

This patch set `embeddedCode` component to be printable by default and provides a method to hide embedded code component on print by detecting the embedded code start with comment as the declaration


nickhsine:
LGTM

PR #1211, opened 2019-06-26
Update with new twreporter-npm-packages

YuCJ:
- [x] Use index components from `@twreporter/index-page` instead of `@twreporter/react-components`
- [x] Use constants and utils from `@twreporter/core` instead of `@twreporter/react-components`
- [x] Update dependencies


nickhsine:
LGTM

PR #1212, opened 2019-06-27
Bug fix: Use `ogImage` rather than `leadingImage` as og image on topic landing page

taylrj:
This patch uses `ogImage` rather than `leadingImage` while keeping `leadingImage` as a default value of og image on topic landing page


nickhsine:
LGTM

PR #1236, opened 2019-07-05
Use redux store to cache settings.fontLevel on browser

nickhsine:
### PR Dependencies
- https://github.com/twreporter/react-article-components/pull/30
- https://github.com/twreporter/twreporter-redux/pull/31

Please see the above PRs before this one.

### DIFF
--- a/src/containers/Article.js
+++ b/src/containers/Article.js
@@ -45,8 +45,12 @@ const _ = {
   get
 }
 
-const { actions, reduxStateFields, utils } = twreporterRedux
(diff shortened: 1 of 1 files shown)


src/containers/Article.js
```diff
@@ -121,13 +125,11 @@ class Article extends PureComponent {
   constructor(props, context) {
     super(props, context)
     this.state = {
-      fontSize:'medium',
-      isFontSizeSet:false
+      fontLevel: props.fontLevel
```
YuCJ: Do we still need `state.fontLevel` here? Or can we just get the value from props given by redux?

PR #1301, opened 2019-07-31
bug fix: background-color of app-shell is not correct while routing

nickhsine:


YuCJ:
Fix the problem with #1302

PR #1302, opened 2019-07-31
Fix worng seleted-page gived to uiManager

YuCJ:
**Problem**: When route changed, the `render` prop of `react-router` will be triggered before the 'CHANGE_SELECTED_POST` action being dispatched. The action is to update the current post. So it will get the wrong `post` data to build the layout.

**Solution**: Change `AppShell` to a redux container in order to update the layout when `selectedPost` in the redux store changed.


nickhsine:
LGTM
Thanks!

PR #1308, opened 2019-08-02
Improve docker image

babygoat:


babygoat:
@nickhsine @YuCJ @taylrj  
I made some changes to the CI  and docker image construction flows, which should reduce the size during the deployment and speed up the building flow.

For some quick validations, the deployment time can be saved more than 50% and size could be reduced to 25% with respect to the original status.

.dockerignore
```diff
@@ -0,0 +1,19 @@
...
+*.md
+*.txt
+*.png
+Makefile
+package.json
+build-babel-preset.js
+build-pm2-processes.js
+yarn.lock
+
+# Binaries
```
nickhsine: remove `# Binaries` ?

.circleci/config.yml
```diff
@@ -15,17 +15,31 @@ jobs:
...
 
       - save_cache:
           paths:
             - node_modules
-          key: v1-dependencies-{{ checksum "yarn.lock" }}
+          key: v1-dependencies-build-{{ checksum "yarn.lock" }}
 
       - run: make build RELEASE_BRANCH=${CIRCLE_BRANCH}
 
+      - run: rm -rf node_modules
```
nickhsine: Is it ok to remove `node_modules` folder?

babygoat: The packages for runtime are installed in the following steps.

nickhsine:
LGTM

.dockerignore
babygoat: This should be executables. But I was wondering why these two files are executable.

babygoat:
updated

PR #1309, opened 2019-08-02
Disable service-worker on preview branch

babygoat:
This patch disables the service-worker on preview branch so that the
basic authentication could work as expected.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -89,7 +89,7 @@ twreporterRedux.createStore(reduxState, '', __DEVELOPMENT__)
 /* eslint-env browser */
 'use strict'
 
-if ('serviceWorker' in navigator && !__DEVELOPMENT__) {
+if ('serviceWorker' in navigator && !__DEVELOPMENT__ && releaseBranch !== 'preview' ) {
   // Delay registration until after the page has loaded, to ensure that our
   // precaching requests don't degrade the first visit experience.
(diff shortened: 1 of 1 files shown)


babygoat:
We still need to clear the indexDB(sw.js) in the client side after this patch. After discussion with @YuCJ , for the time being, clear the storages for those who are required to use preview pages.

src/client.js
```diff
@@ -89,7 +89,7 @@ twreporterRedux.createStore(reduxState, '', __DEVELOPMENT__)
 /* eslint-env browser */
 'use strict'
 
-if ('serviceWorker' in navigator && !__DEVELOPMENT__) {
```
nickhsine: use `src/constants/release-branch` ?

nickhsine:
LGTM

src/client.js
```diff
@@ -89,7 +89,7 @@ twreporterRedux.createStore(reduxState, '', __DEVELOPMENT__)
...
+if ('serviceWorker' in navigator && !__DEVELOPMENT__ && releaseBranch !== 'preview' ) {
```
YuCJ:
```js
import releaseBranchConsts from '@twreporter/core/lib/constants/release-branch'

 if ('serviceWorker' in navigator && !__DEVELOPMENT__ && releaseBranch !== releaseBranchConsts.preview ) { 
```

babygoat:
Updated.

Issue #1110, opened 2019-02-26
Wrong paging indicators of Members section in about-us page


nickhsine:
### Problem Description
In Members(02) section of [about-us page](https://www.twreporter.org/about-us), the paging indicators  will be wrong after resizing the window. It causes the empty paging results.

nickhsine:
I will also create a sub-item in the task card of bugs.
We can fix this when we are available.

PR #1363, opened 2019-09-11
Bug fixes for about us page

taylrj:
Adress #1110 

[Kanbanflow Card](https://kanbanflow.com/t/2GTmSZms)


src/components/about-us/utils/lottie.js
```diff
@@ -29,15 +29,17 @@ export default class Lottie extends PureComponent {
...
     /* Recreate the animation handle if the data is changed */
     if (this.options.animationData !== nextProps.options.animationData) {
       this.deRegisterEvents(this.props.eventListeners)
       this.destroy()
       this.options = { ...this.options, ...nextProps.options }
       this.anim = lottie.loadAnimation(this.options)
       this.registerEvents(nextProps.eventListeners)
+      return true
     }
+    return false
```
nickhsine: I think `shouldComponentUpdate` is used to check the differences between `this.(props|state)` and `(next(Props|State))`.
Should we avoid doing something like `Ln.35 -> Ln.39`?

taylrj: Updated

src/components/about-us/utils/lottie.js
```diff
@@ -29,19 +29,21 @@ export default class Lottie extends React.Component {
...
-      this.registerEvents(this.props.eventListeners)
+      this.registerEvents(nextProps.eventListeners)
+      return true
     }
-    return null 
+    return false
   }
 
   componentDidUpdate() {
+    this.options = { ...this.options, ...this.props.options }
```
taylrj: Updated.

In this line, `this.options` and `this.props.option` need to be merged together as the updated `this.options` since `this.options` contains local variables which are not existed in `this.props.options`.

taylrj:
Updated.

nickhsine:
LGTM

PR #1438, opened 2019-10-15
Use node directorly to run server

babygoat:
This patch removes pm2 and use docker as well kubernetes to manage the
process.


Dockerfile
```diff
@@ -1,18 +1,19 @@
...
+RUN chown ${app_user}:${app_group} /home/${app_user}/src
+
+WORKDIR /home/${app_user}/src
 
 # Add built source files
 COPY . ./
 
 EXPOSE 3000
-CMD pm2 start --no-daemon processes.json
+CMD node dist/server.js
```
nickhsine: Lacks of `NODE_ENV` and `RELEASE_BRANCH` here.
We might need to discuss how to pass environment variables into `process.env` if we drop `pm2` and `processes.json`.

babygoat:
Updated.

PR #1465, opened 2019-10-28
Fix list isFethcing logic

YuCJ:
If there’s only one item in a list, the current code will regard it to be at `isFetching` status.


YuCJ:
Merged into #1466

PR #1476, opened 2019-10-30
[Bug] Fix `mobile-slide-down-menu` where `議題` button doesn't scroll to right position

taylrj:
[KanbanFlow Card](https://kanbanflow.com/t/H9a4fnka)

### DIFF
--- a/src/utils/hash-link-scroll.js
+++ b/src/utils/hash-link-scroll.js
@@ -1,0 +1,62 @@
+let observer = null
+let asyncTimerId = null
+
+function reset() {
+  if (observer) {
+    observer.disconnect()
+  }
+  if (asyncTimerId) {
+    clearTimeout(asyncTimerId)
+  }
+}
+
+function getElementAndScroll(id) {
+  const element = document.getElementById(id)
+  if (element) {
+    element.scrollIntoView()
+    reset()
+    return true
+  }
+  return false
+}
+
+// handle hash link scroll
+// scroll to the anchor after DOM rendered
+function hashLinkScroll() {
+  if (window) {
+    const { hash } = window.location
+    if (hash !== '') {
+      // Push onto callback queue so it runs after the DOM is updated
+      setTimeout(() => {
+        const id = hash.replace('#', '')
+        // since id will be encoded by browser,
+        // decode it back as a string
+        const decodedID = decodeURIComponent(id)
+        // Create a `MutationObserver` instance to watch for changes being made to the DOM tree.
+        // Since `document.getElementById(id)` returns null when navigating from a different page 
+        // without element of id, for example: When navigating from an article page to homepage
+        // with hash `#categories` in url, `document.getElementById('categories')` returns null.
+        // In such case, the `MutationObserver` instance keeps watching and will call it's callback 
+        // when the DOM tree changes, page scrolls to the desired element as long as it can be gotten by id.
+        if (!getElementAndScroll(decodedID)) {
+          if (!observer) {
+            observer = new MutationObserver(() => { getElementAndScroll(decodedID) })
+          }
+          observer.observe(document, {
+            attributes: true,
+            childList: true,
+            subtree: true
+          })
+          // if the element doesn't show up in 10 seconds, stop checking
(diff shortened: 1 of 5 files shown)


src/utils/hash-link-scroll.js
```diff
@@ -0,0 +1,62 @@
...
+        // In such case, the `MutationObserver` instance keeps watching and will call it's callback 
+        // when the DOM tree changes, page scrolls to the desired element as long as it can be gotten by id.
+        if (!getElementAndScroll(decodedID)) {
+          if (!observer) {
+            observer = new MutationObserver(() => { getElementAndScroll(decodedID) })
+          }
+          observer.observe(document, {
+            attributes: true,
+            childList: true,
+            subtree: true
```
nickhsine: `
NOTE: At the very least, childList, attributes, or characterData must be set to true. Otherwise, "An invalid or illegal string was specified" error is thrown.
`
ref: https://developer.mozilla.org/zh-TW/docs/Web/API/MutationObserver

src/utils/hash-link-scroll.js
```diff
@@ -0,0 +1,62 @@
...
+        // Create a `MutationObserver` instance to watch for changes being made to the DOM tree.
+        // Since `document.getElementById(id)` returns null when navigating from a different page 
+        // without element of id, for example: When navigating from an article page to homepage
+        // with hash `#categories` in url, `document.getElementById('categories')` returns null.
+        // In such case, the `MutationObserver` instance keeps watching and will call it's callback 
+        // when the DOM tree changes, page scrolls to the desired element as long as it can be gotten by id.
+        if (!getElementAndScroll(decodedID)) {
+          if (!observer) {
+            observer = new MutationObserver(() => { getElementAndScroll(decodedID) })
+          }
```
nickhsine: Sine `hash-link-scroll.js` is exported as a module,
every client importing it has the same `observer` and `asyncTimerId ` closure(refer to the same ones).

Should we check `observer` is defined or not here?
or just create an observer every time?

BTW, `asyncTimeId` is a closure, would it not be cleared if clients jump to another url with hash param?

nickhsine: I think `observer` and `asyncTimeId` should be moved into `hashLinkScroll` function as  local variables, not closures of the `hash-link-scroll` module.

taylrj: Thanks, it has been updated in commit `52b5b89`.

src/utils/hash-link-scroll.js
```diff
@@ -28,28 +35,29 @@ function hashLinkScroll() {
...
         // Since `document.getElementById(id)` returns null when navigating from a different page 
         // without element of id, for example: When navigating from an article page to homepage
         // with hash `#categories` in url, `document.getElementById('categories')` returns null.
         // In such case, the `MutationObserver` instance keeps watching and will call it's callback 
         // when the DOM tree changes, page scrolls to the desired element as long as it can be gotten by id.
-        if (!getElementAndScroll(decodedID)) {
-          if (!observer) {
-            observer = new MutationObserver(() => { getElementAndScroll(decodedID) })
-          }
+        if (!getElementAndScroll(decodedId)) {
```
nickhsine: `getElementAndScroll` -> `getElement` ?

src/utils/hash-link-scroll.js
```diff
@@ -28,28 +35,29 @@ function hashLinkScroll() {
...
+          observer = new MutationObserver(() => { getElementAndScroll(decodedId, observer, asyncTimerId) })
```
nickhsine: `getElementAndScroll ` -> `scrollToElement(element, observer, asyncTimeId)` ?

nickhsine: Maybe we could remove `scrollToElement` function or rename it?
After second thoughts, `observer` and `asyncTimeId` is not necessary for `scrollToElement`.

taylrj: That's great!!
I move out `reset` function call so as not to send `observer` and `asyncTimeId` as params which are not necessary for `scrollToElement` indeed.

Please see commit `2d77509 `

PR #1478, opened 2019-11-04
[Bug] Can't scroll to correct anchor on index page

taylrj:
[Kanban Flow Card](https://kanbanflow.com/t/H9a4fnka)

### DIFF
--- a/src/utils/hash-link-scroll.js
+++ b/src/utils/hash-link-scroll.js
@@ -13,13 +13,14 @@ function reset(observer, asyncTimerId) {
 }
 
 /*
- * @param {string} id - element id 
+ * @param {string} id - element id
+ * @param {number} delay - delay time to wait until other effect ends
  * @returns {boolean} - represents whether document can get element by id 
  */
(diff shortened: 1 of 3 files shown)


src/utils/hash-link-scroll.js
```diff
@@ -19,7 +19,7 @@ function reset(observer, asyncTimerId) {
 function scrollToElement(id) {
   const element = document.getElementById(id)
   if (element) {
-    element.scrollIntoView()
+    setTimeout(() => { element.scrollIntoView() }, 200)
```
YuCJ: Set the delay by function parameter or outside the function to make the delay time more flexiable? Example useage:

```js
const id = 'element-id'
const delay = 200  // Maybe the duration time of other effect. Wait until it's end.
scrollToElement(id, delay)
```

or


```js
const id = 'element-id'
const delay = 200
setTimeout(() => { scrollToElement(id) }, delay)
```

taylrj: Updated.

PR #1498, opened 2019-11-07
[hotfix] Fix og-image type error on topic landing page

taylrj:
This patch fixes og-image type error where should be an url string
rather than `predefinedPropTypes.imgObj`


src/containers/TopicLandingPage.js
```diff
@@ -89,6 +89,7 @@ class TopicLandingPage extends Component {
     const teamDescription = _.get(topic, 'team_description.api_data', [])
     const ogDescription =  _.get(topic, 'og_description', '') || SITE_META.DESC
     const ogTitle = _.get(topic, 'og_title', '') || _.get(topic, 'title', '')
+    const ogImageUrl = _.get(topic, 'og_image.resized_targets.tablet.url')
```
YuCJ: We can add `|| _.get(topic, 'leading_image.resized_targets.tablet.url')` for safe

taylrj: Updated.

PR #1536, opened 2019-12-27
refactor: specify stackdriver resource object

nickhsine:
### KanbanFlow Card
https://kanbanflow.com/t/gi1gCf6C

### DIFF
--- a/src/components/ErrorBoundary.js
+++ b/src/components/ErrorBoundary.js
@@ -13,7 +13,10 @@ class ErrorBoundary extends React.Component {
   }
 
   componentDidCatch(error, info) {
-    logger.error('An error was catched in React ErrorBoundary, ', error)
+    logger.errorReport({
+      message: 'An error was catched in React ErrorBoundary.',
+      report: error
+    })
     this.setState({ hasError: true })
     this._error = error
     this._info = info
--- a/src/components/Pagination.js
+++ b/src/components/Pagination.js
@@ -27,7 +27,10 @@ class EnhancedPagination extends PureComponent {
     try {
       this.props.history.push(to)
     } catch (err) {
-      logger.error(`Error on history.push(${to}), `, err)
+      logger.errorReport({
+        report:err,
+        message: `Error on history.push(${to}).`
+      })
     }
   }
 
--- a/src/data-loaders/bookmark-list.js
+++ b/src/data-loaders/bookmark-list.js
@@ -29,8 +29,11 @@ export default function loadData({ store }) {
     const jwt = _.get(state, [ reduxStatePropKey.auth, 'accessToken' ])
     const userID = _.get(state, [ reduxStatePropKey.auth, 'userInfo', 'user_id' ])
     return store.dispatch(getMultipleBookmarks(jwt, userID, 0, defaultLimit, defaultSort))
-      .catch(error => {
-        logger.error('Bookmark list data loader can not load data, ', error)
+      .catch(err => {
+        logger.errorReport({
+          report: err,
+          message: 'Bookmark list data loader can not load data.'
+        })
       })
   }
   return Promise.resolve()
--- a/src/express/server.js
+++ b/src/express/server.js
@@ -169,7 +169,11 @@ class ExpressServer {
         return next(err)
       }
 
-      logger.error(err)
+      logger.errorReport({
+        report: err,
+        message: 'Express server handles unexpected error.'
+      })
+
       res.header('Cache-Control', 'no-store')
       if (err instanceof NotFoundError || _.get(err, 'response.status') === 404) {
         res.redirect('/error/404')
@@ -184,8 +188,10 @@ class ExpressServer {
       const mw = await loggerFactory.makeExpressMiddleware()
       this.app.use(mw)
     } catch(err) {
-      logger.error('Cannot apply logger express middleware')
-      logger.error(err)
+      logger.errorReport({
+        report: err,
+        message: 'Express cannot apply logger middleware.'
+      })
     }
   }
 
@@ -219,12 +225,18 @@ class ExpressServer {
     if (port) {
       this.server.listen(port, (err) => {
         if (err) {
-          logger.error(err)
+          logger.errorReport({
+            report: err,
+            message: `Express server cannot listen on host ${host} and port ${port}.`
+          })
+          return
         }
         logger.info('==> 💻  Open http://%s:%s in a browser to view the app.', host, port)
       })
     } else {
-      logger.error('==>     ERROR: No PORT environment variable has been specified')
+      logger.errorReport({
+        message: '==>     ERROR: No PORT environment variable has been specified.'
+      })
     }
   }
 }
--- a/src/logger/index.js
+++ b/src/logger/index.js
@@ -5,21 +5,50 @@ function loggerFactory() {
   let logger
 
   function create() {
-    // on server side,
-    // we use winston and logging-winston to log
-    if (!globalEnv.isBrowser) {
+    let logger
 
-      if (globalEnv.isProduction) {
-        return createProdLogger()
-      }
-      return createDevLogger()
+    if (globalEnv.isBrowser) {
+      // on client side,
+      // we use console directly
+      // TODO:
+      // send client side log to stackdriver logging
+      logger = console
+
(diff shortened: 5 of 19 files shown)


src/logger/winston-logger.js
```diff
@@ -1,11 +1,38 @@
...
+/**
+ *  @param {string} releaseBranch - 'master', 'preview', 'staging' or 'release'
+ *  @return {string} kubernetes namespace
+ */
+function getK8sNamespaceName(releaseBranch) {
+  switch(releaseBranch) {
+    case 'release':
+      return 'production'
+    default:
+      return releaseBranch
```
babygoat: Is preview, master namespace available?

nickhsine: Yes, they are available on stackdriver logging.
However, in theory, `master` release branch won't exist on production environment .

babygoat: I don't think the preview namespace is available. In keystone-cluster, only `default` namespace is used?

nickhsine: It's a good question.
I think I should move `preview` release branch logs to `keystone-cluster`.

src/logger/winston-logger.js
```diff
@@ -1,11 +1,38 @@
...
+      type: 'k8s_container',
+      labels: {
+        cluster_name: 'twreporter',
+        namespace_name: getK8sNamespaceName(globalEnv.releaseBranch),
+        container_name: 'twreporter-website-v2'
+      }
+    },
     // for stackdriver error reporting
     serviceContext: {
       service: 'twreporter-react-server',
```
babygoat: Why does the serviceContext.version not match with the service version?

nickhsine: Which service version do you mean? The package.json#version?

src/logger/winston-logger.js
babygoat: BTW, keystone-cluster currently stays on kubernetes1.12 which does not offer the latest stackdriver logging mechanism. If our plan does not include the keystone-cluster migration, this will block the preview branch from release.

nickhsine: I will use default resource type (legacy stackdriver used) for `preview` release branch.

src/logger/winston-logger.js
babygoat: Yes

nickhsine: I will remove `serviceContext.version` here.

babygoat: I'm not sure if the entry removed will break the serviceContext object or not. I'm not objecting to specify the version but to make the meaning of the version clear.
https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext

nickhsine: it won't break  `serverContext` object since `serviceContext.version` is optional (https://github.com/googleapis/nodejs-logging-winston/blob/master/src/types/core.d.ts#L127).

src/logger/winston-logger.js
nickhsine: > BTW, keystone-cluster currently stays on kubernetes1.12 which does not offer the latest stackdriver logging mechanism. If our plan does not include the keystone-cluster migration, this will block the preview branch from release. I plan to have the migration to GKE 1.14 with private cluster setup and Identity aware proxy.

Sure, we can plan the migration.
After migration, we can update the `resource` object here to upload the logs with new stackdriver logging API.

src/logger/winston-logger.js
babygoat: The interface declaration stands for the object for stackdriver logging. The service context entry is not a must. But for errorReporting event, the serviceContext does.

babygoat: I believe it won't break. In other setup, only `service: gke_instance`  is specified

nickhsine: Just test the error reporting without `serviceContext.version`, and it reports error.
However, I plan to pass `globalEnv.pkgVersion` and use it as `serviceContext.version`.

nickhsine:
@babygoat @taylrj 
Please help to review the latest three commits.

src/logger/winston-logger.js
```diff
@@ -22,7 +22,8 @@ export function createProdLogger() {
   const opts = {
     // for stackdriver error reporting
     serviceContext: {
-      service: 'twreporter-react-server'
+      service: 'twreporter-react-server',
```
babygoat: Minor. Why do we want to have different service name  other than container name?

nickhsine:
Updated.

src/logger/winston-logger.js
nickhsine: How about `{cluster_name}:{container_name}:{namespace}:server` ?

src/components/ErrorBoundary.js
```diff
@@ -13,7 +13,8 @@ class ErrorBoundary extends React.Component {
   }
 
   componentDidCatch(error, info) {
-    logger.error('An error was catched in React ErrorBoundary, ', error)
+    logger.error(new Error('An error was catched in React ErrorBoundary. '))
+    logger.error(error)
```
babygoat: Could we have the composite log error message instead of two? This would result two error entries instead of one though only one error occurs.

src/components/Pagination.js
```diff
@@ -27,7 +27,8 @@ class EnhancedPagination extends PureComponent {
     try {
       this.props.history.push(to)
     } catch (err) {
-      logger.error(`Error on history.push(${to}), `, err)
+      logger.error(new Error(`Error on history.push(${to}). `))
+      logger.error(err)
```
babygoat: Same above

src/data-loaders/bookmark-list.js
```diff
@@ -29,8 +29,9 @@ export default function loadData({ store }) {
     const jwt = _.get(state, [ reduxStatePropKey.auth, 'accessToken' ])
     const userID = _.get(state, [ reduxStatePropKey.auth, 'userInfo', 'user_id' ])
     return store.dispatch(getMultipleBookmarks(jwt, userID, 0, defaultLimit, defaultSort))
-      .catch(error => {
-        logger.error('Bookmark list data loader can not load data, ', error)
+      .catch(err => {
+        logger.error(new Error('Bookmark list data loader can not load data.'))
+        logger.error(err)
```
babygoat: Same above.

src/express/server.js
```diff
@@ -169,7 +169,9 @@ class ExpressServer {
         return next(err)
       }
 
+      logger.error(new Error('Express server handles unexpected error. '))
```
babygoat: Same above.

src/components/ErrorBoundary.js
nickhsine: It's on purpose. 
After discussion with @taylrj @YuCJ , we compromise on this error logging convention.
We can try this version for now, and observe the occurrence of error logs.

nickhsine: The reason I wrote two error logs here is because of error reporting.
If `error` is not an `Error` object, then the error reporting won't catch the error.
Instead of checking `error instanceof Error` for every `logger.error`,
I plan to write informative error message for logging (since `error` comes everywhere, we need to do lots of check to know what it is).

babygoat: Why and what do we want to have the compromise for?

nickhsine:
Updated. @babygoat @YuCJ @taylrj  please have a see.
I compact two `logger.error`s into one.
And most important of all, I reserve the error (which could be an `Error` instance, string or object) as `jsonPayload.metadata.__errorReport` on stackdriver logging viewer.

src/logger/index.js
```diff
@@ -5,21 +5,49 @@ function loggerFactory() {
...
+    }
+
+
     // on server side,
     // we use winston and logging-winston to log
-    if (!globalEnv.isBrowser) {
+    if (globalEnv.isProduction) {
+      logger = createProdLogger()
+    }
+    logger = createDevLogger()
```
babygoat: Create ProdLogger() if isProduction is true?

nickhsine: It's a bug.
updated.

babygoat:
LGTM

taylrj:
LGTM

PR #1560, opened 2020-01-16
Handle conditional requests

nickhsine:
### Change Reason
Even though we set `Cache-Control: public, max-age:300` in response header,
the service worker would use **Network falling back to cache** policy (see https://github.com/twreporter/twreporter-react/blob/master/service-worker/service-worker.tmpl#L42-L50) to fetch the resource from network, and fallback to browser cache if request fails.

This policy will make server facing much more requests than we thought.
Therefore, the server needs to handle conditional requests.
Conditional requests make server not to send body payload back to browsers if the response body is the same as the last time response. 

### Background Information
The reason service worker uses **Network falling back to cache policy** is because we want to reflect the resource change as soon as possible.
We can have another thread to discuss what kind of policy we should use. (See https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook#serving-suggestions)

### DIFF
--- a/src/constants/status-code.js
+++ b/src/constants/status-code.js
@@ -1,6 +1,7 @@
 export default {
   ok: 200,
   created: 201,
+  movedPermanently: 301,
   badRequest: 400,
   unauthorized: 401,
   forbidden: 403,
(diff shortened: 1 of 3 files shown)


nickhsine:
### Reason to Close PR
Express set weak `Etag` response header by default.
Therefore, we already handle conditional requests implicitly when we start up the express server.

However, the responses browser get are without `Etag` header.
After digging out, `Etag` response header is removed by Cloudflare.
'A Better Browser', which is a Cloudflare App, is installed on our Cloudflare,
and it removes `Etag` response header.
After uninstalling that Cloudflare App, the `Etag` is showing up, and the Conditional Requests works perfectly.

PR #1562, opened 2020-02-06
fix: log out the user if fail to get access token

YuCJ:
If the user is authenticated but failed to get access token, sign the user out (clear her/his id token).


nickhsine:
LGTM

PR #1610, opened 2020-03-31
fix: section 2 of about us

taylrj:
This patch fixes section2 where arrows of photography department can not
click


nickhsine:
LGTM

PR #1629, opened 2020-04-24
Release v4.4.22 on preview server

taylrj:
This PR is to trigger CircleCI to build and deploy v4.4.22 on preview server.


nickhsine:
This PR is too big to review.
I think we can merge and release it to test.
But before releasing it, we may need to notify editors that we are going to upgrade preview server,
there will be a down time for preview server. 
Make sure that they have time to prepare it.

PR #1644, opened 2020-06-12
Update about-us

taylrj:


nickhsine:
I suggest that we could separate the content from the code base.
There will be several advantages.

1. we don't need to release a new version of `twreporter-react`.
2. we don't need to rebuild the webpack bundles, so the cache won't be invalid.
3. we can let editors to do the routine tasks.

How do you think?

taylrj:
Yes, that would be great and is what we need to work on.

Also, I think this is an independent story, maybe need some time to make it happen.
Since the requirement of this change has a due date (around next Monday) which has been confirmed with marketing manager, I will work on [this story](https://kanbanflow.com/t/vPBTUGJ1) after this version is deployed to production.

How do you think ?

nickhsine:
Sure, we can release version since we have time pressure.

PR #1687, opened 2020-08-28
refactor: use memoize-one to avoid re-rendering

nickhsine:
In this patch, we use memoize-one to avoid
re-shallowCloning (fullTopic|fullPost).

It helps to avoid re-rendering article page body
and topic landing page body if end users just click load more related
articles button.


nickhsine:
@YuCJ 
This PR is a hot fix.
I merge it first,
if there is anything to change, please feel free to leave the comments.

PR #1718, opened 2020-09-14
refactor: do not cache response for preview branch and update service worker runtime cache pattern

nickhsine:
### Change Reason
Editors report that preview server (keystone-preview.twreporter.org) didn't show the latest content after modifying it in the keystone-editor.

After digging out, I guess the reason is the content is cached by browser.
Therefore, we remove `Cache-Control: public, max-age=300` response header for preview server.

### DIFF
--- a/src/express/server.js
+++ b/src/express/server.js
@@ -8,10 +8,11 @@ import get from 'lodash/get'
 import globalEnv from '../global-env'
 import http from 'http'
 import initReduxStoreMiddleware from './middlewares/init-redux-store'
-import loggerFactory from '../logger'
 import path from 'path'
+import loggerFactory from '../logger'
 import releaseBranchConsts from '@twreporter/core/lib/constants/release-branch'
 import renderHTMLMiddleware from './middlewares/render-html'
+import statusCodeConst from '../constants/status-code'
 
 const _ = {
   get,
@@ -93,7 +94,7 @@ class ExpressServer {
         res.format({
           'text/plain': function() {
             res
-              .status(200)
+              .status(statusCodeConst.ok)
               .send(
                 'User-agent: * \n' +
                   'Sitemap: https://public.twreporter.org/sitemaps/www-sitemap.xml\n' +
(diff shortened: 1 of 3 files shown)


taylrj:
LGTM

src/express/server.js
```diff
@@ -174,7 +174,10 @@ class ExpressServer {
         const statusCode = _.get(routerStaticContext, 'statusCode', statusOK)
         if (!res.headersSent && statusCode < statusRedirect) {
           const idToken = _.get(req, 'cookies.id_token')
-          if (idToken) {
+          if (
+            idToken ||
+            globalEnv.releaseBranch === releaseBranchConsts.preview
```
babygoat, line 179: The logic here is getting complicate. I suggest we should extract the logic here to a separate function like `IsCacheable()`. Remove the comment `//not to cache personal response` below. Leave the comments to indicate in what situation the cache does not apply.

nickhsine:
Updated.

src/express/server.js
```diff
@@ -160,32 +163,48 @@ class ExpressServer {
...
+        const statusCode = _.get(
+          routerStaticContext,
+          'statusCode',
+          statusCodeConst.ok
+        )
+        res.locals.statusCode = statusCode
+        next()
+      },
+      function setResponseHeader(req, res, next) {
+        if (globalEnv.releaseBranch === releaseBranchConsts.preview) {
```
babygoat, line 186: Leave the comment here to describe why we disable cache on preview branch

babygoat:
Also, minor suggestion, split the statusCodeConst changes with the function extract patch for better review.

src/express/server.js
nickhsine: Oops. Forget to add comments.

nickhsine:
Updated.

babygoat:
LGTM

nickhsine:
Updated according to your suggestion.

babygoat:
LGTM

Issue #150, opened 2016-02-03
Footer style is inconsistent


sorarize:
首頁
![image](https://cloud.githubusercontent.com/assets/240043/12773581/fb7d4306-ca77-11e5-83f5-293176be22d9.png)

[專欄](https://www.twreporter.org/a/opinion-japan-report-twelection)
![image](https://cloud.githubusercontent.com/assets/240043/12773585/0b9b714a-ca78-11e5-833d-c77316bcef2a.png)

[infographic](https://www.twreporter.org/a/infographics-2016-election)
![image](https://cloud.githubusercontent.com/assets/240043/12773657/a26bef0a-ca78-11e5-9002-407161030084.png)

kidwm:
BTW, how about move `twreporter-react-footer-components` and other components to npm orgs?

https://www.npmjs.com/docs/orgs/

nickhsine:
The footer is consistent now.

nickhsine:
@kidwm thanks for your suggestion, we will try it.

PR #1722, opened 2020-09-15
fix: render stale data due to wrong memoize function

nickhsine:
### Change Reason
In memoize function,
we check `prevPost.id === nextPost.id` and
`prevPost.full === nextPost.full`,
if true, memoize will use the cached post object.

However, that condition makes memoize to return the stale post object.
For example, if editor edits the post object, such as changing the post title,
that change won't make the memoize function to return a new post object.
Therefore, Article page will render the stale data.

### Related PR
[refactor: stop using lodash.merge in entities reducer #150
](https://github.com/twreporter/twreporter-npm-packages/pull/150)

The reason why we use `arePostsIdAndFullEqual` and `areTopicsIdAndFullEqual` is
`redux.entities.posts` and `redux.entities.topics` change all the time.
And that change will cause Article page re-render.
In the [above PR](https://github.com/twreporter/twreporter-npm-packages/pull/150), it stops changing `redux.entities.posts` and `redux.entities.topics` if there is nothing new.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "@twreporter/index-page": "1.0.9",
     "@twreporter/react-article-components": "^1.2.3",
     "@twreporter/react-components": "^8.3.1",
-    "@twreporter/redux": "^7.0.1",
+    "@twreporter/redux": "7.0.2-rc.0",
     "@twreporter/universal-header": "^2.1.6",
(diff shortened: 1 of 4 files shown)


taylrj:
LGTM 👍

PR #1727, opened 2020-10-06
feat: use `og_image` as image fallback for topic landing page

taylrj:
Use `og_image` as image fallback for related items of topic landing page


nickhsine:
LGTM

PR #1738, opened 2020-10-19
fix: handle catId not existed. return 404 not found res

nickhsine:
### Reason to Change
In our [error reporting system](https://console.cloud.google.com/errors/COTC6tOXo6DYPQ?time=P30D&project=coastal-run-106202), there are lots of errors due to `catId` not existed.
However, that error should not be an error to be reported.
That error should be 404 (NotFound) error, not 500 (Internal server error) error.


taylrj:
LGTM

PR #1739, opened 2020-10-20
fix: handle other errors except for 404 and 500 errors

nickhsine:
### Reason to Change
In our [error reporting system](https://console.cloud.google.com/errors/CPLAmLXIk_eGPg?time=P30D&project=coastal-run-106202), there are lots of `Not Acceptable` error.
However, `Not Acceptable` error (status code: 406), which is a client error, should not be reported.

### Express Custom Error Misbehavior
In our old logic, if we encounter 406 error, we will redirect that error to 500 error page.
Therefore, error reporting system will be triggered and the worst part is we return the wrong response (response should be 406, rather than 500).

In the new way, we not only handle 404 and 500 error (redirect to error page), but also 
log error information if error.statusCode < 500, and report error if error.statusCode >= 500.
However, since our error page only handle 404 and 500 error, we need to refactor `@twreporter/react-components/lib/error` to handle other error codes.
Before that refactoring, we simply response error code.

### More About Not Acceptable Error
In our `/robots.txt` route, we response the content with `Content-Type: text/plain` header.
However, some requests for `/robots.txt` contains `Accept` header, and that `Accept` header does not have `text/plain` type. That makes express create 406 Not Acceptable error.
We should return 406 Not Acceptable response to that requests, rather than 500 error page.

### DIFF
--- a/src/express/server.js
+++ b/src/express/server.js
@@ -22,7 +22,7 @@ const logger = loggerFactory.getLogger()
 
 class ExpressServer {
   constructor() {
-    this.app = new Express()
+    this.app = Express()
     this.server = new http.Server(this.app)
   }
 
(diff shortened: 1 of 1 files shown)


src/express/server.js
```diff
@@ -224,15 +224,46 @@ class ExpressServer {
...
+            // server error report
+            logger.errorReport({
+              report: err,
+              message:
+                'Server error (but not internal server error) was caught by Express custom error handler.',
+            })
+          }
+
+          // TODO: render error page with follow-up guidance to end users
+          res.sendStatus(errStatusCode)
```
taylrj, line 265: What would be the follow-up guidance ?

src/express/server.js
```diff
@@ -224,15 +224,46 @@ class ExpressServer {
...
+          } else {
+            // server error report
+            logger.errorReport({
+              report: err,
+              message:
+                'Server error (but not internal server error) was caught by Express custom error handler.',
+            })
+          }
+
+          // TODO: render error page with follow-up guidance to end users
```
taylrj, line 264: What would be the follow-up guidance ?

nickhsine: Just like 500 or 404 error page, we have instructions to tell end users what's going on now, and we could direct them to other pages or actions.
How do you think?

taylrj: Yeah, that would be great.

taylrj:
LGTM 👍

PR #1743, opened 2020-10-22
fix: handle author page 404 error

nickhsine:
### Reason to Change
In our error reporting system, we have been getting lots of server errors on author page (see [stackdriver error reporting](https://console.cloud.google.com/errors/CNHgkt_OkYeULg?time=P30D&project=coastal-run-106202)).

The root cause is that some end users requests url like `www.twreporter.org/authors/process064.html` (see [stackdriver logger](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%20resource.labels.cluster_name%3D%22twreporter%22%20resource.labels.namespace_name%3D%22production%22%20resource.labels.container_name%3D%22twreporter-nginx-v2%22%0AinsertId%3D%224nyzyf7ohhriubr1o%22;timeRange=2020-10-21T20:39:00.207Z%2F2020-10-21T20:40:00.207Z?project=coastal-run-106202)).
There is no data for such requests. 
Currently, our system responses 500 error page. 
However, we should response 404 error page instead.

### Solution
Due to [this PR](https://github.com/twreporter/twreporter-npm-packages/pull/155), 
our author page data loader could be updated to
- treat empty payload as 404 error

### DIFF
--- a/src/data-loaders/author-page.js
+++ b/src/data-loaders/author-page.js
@@ -1,5 +1,6 @@
 // lodash
 import get from 'lodash/get'
+import statusCodeConst from '../constants/status-code'
 
 const _ = {
   get,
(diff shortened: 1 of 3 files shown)


src/data-loaders/author-page.js
```diff
@@ -1,5 +1,4 @@
 // lodash
-import actionTypes from '@twreporter/redux/lib/constants/action-types'
 import get from 'lodash/get'
 import set from 'lodash/set'
```
taylrj, line 3: remove unused `lodash/set` ?

taylrj:
LGTM 👍

PR #1744, opened 2020-10-23
refactor: update express server error handling

nickhsine:
We used to delegate the errors from data loaders to the custom error handler.

In custom error handler, we log client error (through stackdriver logging) and report server error (through stackdriver error reporting).

However, errors from data loaders should be treated in different way.

For example, data loaders will request data from backend servers, and that request might cause client
error, like 400 (Bad request).
If we pass 400 error to custom error handler, that error will be only logged, not reported.
We, as developers, need to know why that request causes 400 error.

Therefore, this patch adds another error handler for data loaders middleware.

### DIFF
--- a/src/express/middlewares/data-loader.js
+++ b/src/express/middlewares/data-loader.js
@@ -1,12 +1,16 @@
 /* eslint node/no-deprecated-api: 1 */
 import get from 'lodash/get'
 import getRoutes from '../../routes'
+import loggerFactory from '../../logger'
+import statusCodeConst from '../../constants/status-code'
 import url from 'url'
 import { matchPath } from 'react-router-dom'
 
 const _ = {
   get,
 }
+
+const logger = loggerFactory.getLogger()
 
 /**
  *  This middleware prepare data according to routes
@@ -50,7 +54,22 @@ function dataLoaderMiddleware(namespace) {
         next()
       })
       .catch(failAction => {
-        next(_.get(failAction, 'payload.error'))
+        const error = _.get(failAction, 'payload.error')
+        const errStatusCode =
+          _.get(error, 'statusCode') || _.get(error, 'status')
+
+        switch (errStatusCode) {
+          case statusCodeConst.notFound: {
+            return res.redirect(`/error/${errStatusCode}`)
+          }
+          default: {
+            logger.errorReport({
+              report: error,
+              message: 'Error was caught by data-loader express middleware.',
+            })
+            return res.redirect(`/error/${errStatusCode}`)
+          }
+        }
       })
   }
 }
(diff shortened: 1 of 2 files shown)


taylrj:
Error handling in server can be defined as following procedures:
- run `logger.info` / `logger.errorReport`
- redirect to error page


IMO, if custom error handler could be the only middleware which is responsible for error handling (defined as above), it would be more understandable.

In general case, run `logger.info` only when the `status code < 500` but not `404` and run `logger.errorReport` when `status code >= 500`. 

So, error handling in data loader is a special case compare with the general case.
How about wrapping the error with an object with a type name and just passing it to custom error handler ?

Follow this idea, here is a proposed solution:

In data-loader middleware:

```js
Promise.all(dataLoadingPromises)
      .then(() => {
        next()
      })
      .catch(failAction => {
        next({
          type: 'data-loader-error'
          payload: _.get(failAction, 'payload.error'))
      })
```

And in `server.js`

```js
const errorType = _.get(err, 'type')
if (typeof errorType === 'string') {
  err = _.get(err, payload)
}
const errStatusCode = _.get(err, 'statusCode') || _.get(err, 'status')

switch(errorType) {
  case 'data-loader-error' {
    if (errStatusCode !== statusCodeConst.notFound) {
      logger.errorReport({ ... })
    }
    break
  }

  default {
    if (errStatusCode < 500) {
      if (errStatusCode !== statusCodeConst.notFound) {
        logger.info({...})
      } 
    } else {
      logger.errorReport({ ... })
    }
    break
  }  
}

return res.redirect(`/error/${errStatusCode}`)
```

In this approach, we can centralize the error handling in custom error handler which contains all of the business logic.
Yes, we need to add additional error type definition and add more complexity in the custom error handler, but if we have other business logic to handle error in the future, just name it in a new type and add corresponding handling logic.

 How do you think?

nickhsine:
Thanks for the reply.

I totally agree with your proposal.
Actually, I have exactly the same thought as yours at first.
We could centralize the error handlers in only one function.

However, what changed me to choose current solution is because of Express.
In Express coding convention, we could stop passing requests to next middlewares if we `res.redirect` in the `data-loader` middleware.

Thus, I think we could follow up the Express coding convention to reduce the response time.

nickhsine:
Updated.

taylrj:
LGTM 👍

nickhsine:
After discussion with @taylrj , we agree that 
1. rename `__applyCustomErrorHandler` to `__applyDefaultErrorHandler`: let default error handler to handle those unhandled errors passed from previous middlewares.

2. each middleware can handle known errors by itself; it can decide how to response.

PR #1760, opened 2020-12-02
Update html.js with right rss feed

typebrook:
@taylrj 

Wrong rss feed in html header


taylrj:
LGTM

Thanks, @typebrook !

PR #1792, opened 2021-02-18
refactor: show article number when an author has no article

taylrj:
This patch updates article collection display logic to show article
number 0 when an author has no article.


babygoat:
LGTM

PR #1793, opened 2021-02-19
refactor: tweak data path according to @twreporter/redux updates

taylrj:
https://github.com/twreporter/twreporter-npm-packages/pull/170 is required for this change.

### DIFF
--- a/src/containers/Author.js
+++ b/src/containers/Author.js
@@ -165,7 +165,7 @@ function mapStateToProps(state, ownProps) {
     title: _.get(authorEntity, 'jobTitle') || '',
     image: authorImage,
     mail: _.get(authorEntity, 'email') || '',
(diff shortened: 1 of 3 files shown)


babygoat:
LGTM

src/containers/Author.js
```diff
@@ -165,7 +165,7 @@ function mapStateToProps(state, ownProps) {
     title: _.get(authorEntity, 'jobTitle') || '',
     image: authorImage,
     mail: _.get(authorEntity, 'email') || '',
```
babygoat, line 167: Note that the email might be empty. Previous approach(algolia-v2) appends the default email(contact@twreporter.org). I suggest we move that fallback here rather than transform the data.

taylrj: Updated.

babygoat:
LGTM

PR #1822, opened 2021-04-19
fix: utilize `window.location.href` as the path for GA

taylrj:
This patch utilizes `window.loaction.href` instead of
`window.location.pathname` as the path for Google Analytics to include
search strings.


babygoat:
LGTM

PR #1830, opened 2021-04-22
fix: utilize `window.location.href` when sending pageview

babygoat:
This patch utilizes the `window.location.href` when sending GA pageview
event on client side.


taylrj:
LGTM

PR #1831, opened 2021-04-23
chore: bump version to 4.6.7-rc.0

babygoat:


taylrj:
LGTM

babygoat:
@taylrj  I suggest we should stabilize the commit

Commit generation convention in backend
```
git log --pretty='format:- [[%C(auto)%h](https://github.com/twreporter/twreporter-react/commit/%H)] - %s(%an)' [TAG]...HEAD --no-mergesd
```

babygoat:
I will merge first to test on staging. We can make the final decision before v4.6.7 release.

taylrj:
Currently, front-end repositories (except for monorepo) use [changelog-maker](https://github.com/nodejs/changelog-maker) to generate commits in `CHANGELOG.md`.

The result looks similar to `git log ...` command as above comment. 
Do you suggest we should unify the way generating commits ?

babygoat:
> Currently, front-end repositories (except for monorepo) use [changelog-maker](https://github.com/nodejs/changelog-maker) to generate commits in `CHANGELOG.md`.
> 
> The result looks similar to `git log ...` command as above comment.
> Do you suggest we should unify the way generating commits ?

IMO, it would make the changelog modification and the releaser development easier if we have a unified style. But we don't have to fulfill it in this patch. I could change the format to be consistent with that from changelog-maker. Besides, I strongly suggest removing the merge commit. What do u think?

taylrj:
Sure, I'd like to remove all of the merge commits since they are not necessary and does not provide more information.

PR #1833, opened 2021-04-27
fix: count the pageviews with pathname + query parameters

babygoat:
This patch counts the pageviews with `window.location.pathname` +
`window.location.search` instead of full pathname `window.location.href`.


taylrj:
LGTM

PR #1847, opened 2021-05-05
fix: incorrect anchor position on index page

taylrj:
[story-48][task-1] fix the anchor on index-page

### DIFF
--- a/src/components/side-bar/base-components.js
+++ b/src/components/side-bar/base-components.js
@@ -73,9 +73,9 @@ class Anchors extends React.PureComponent {
     data.forEach(anchorObj => {
       const id = _.get(anchorObj, 'id', '')
       const label = _.get(anchorObj, 'label', '')
+      const show = _.get(anchorObj, 'show', false)
 
-      // id and label are not empty string
-      if (id && label) {
+      if (show) {
         anchorBts.push(
           this._renderAnchor({
             handleClick: handleClickAnchor,
--- a/src/containers/Home.js
+++ b/src/containers/Home.js
@@ -1,13 +1,14 @@
+import { AnchorWrapper as Section } from '@twreporter/react-components/lib/side-bar'
 import { connect } from 'react-redux'
-import categoryConst from '../constants/category'
 import CSSTransition from 'react-transition-group/CSSTransition'
 import Helmet from 'react-helmet'
-import loggerFactory from '../logger'
 import IndexPageComposite from '@twreporter/index-page'
 import LoadingSpinner from '../components/Spinner'
 import PropTypes from 'prop-types'
-import qs from 'qs'
 import React from 'react'
+import categoryConst from '../constants/category'
+import loggerFactory from '../logger'
+import qs from 'qs'
 import sideBarFactory from '../components/side-bar/side-bar-factory'
 import siteMeta from '../constants/site-meta'
 import styled, { css } from 'styled-components'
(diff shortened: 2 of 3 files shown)


src/components/side-bar/with-anchors.js
```diff
@@ -0,0 +1,113 @@
...
+AnchorWrapper.propTypes = {
+  anchorId: PropTypes.string.isRequired,
+  anchorLabel: PropTypes.string.isRequired,
+  children: PropTypes.element.isRequired,
+}
+
+/**
+ * `withAnchors`
+ * returns a High Order Component setting the anchors onto `DecoratedComponent`.
+ * It passes props `anchors` to the `DecoratedComponent` which subsequently adds the side bar functionalities to its children.
```
babygoat, line 35: sidebar functionality?

src/components/side-bar/with-anchors.js
```diff
@@ -0,0 +1,113 @@
...
+ *      <AnchorWrapper anchorId="section-3" anchorLabel="Section Three">
+ *        <SectionThree />
+ *      </AnchorWrapper>
+ *    </Sidebar>
+ *  )
+ * }
+ *```
+ *
+ * The above codes will get `anchorId`, `anchorLabel` through <AnchorWrapper /> props and generates
+ * the anchors array which is the prop required by SibeBarHOC as shown below.
```
babygoat, line 64: SideBarHOC

src/containers/Home.js
```diff
@@ -295,39 +255,61 @@ class Homepage extends React.PureComponent {
...
+            </Background>
+          </Section>
+          <Section anchorId="infographic" anchorLabel="多媒體">
+            <Background backgroundColor={moduleBackgounds.infographic}>
+              <InforgraphicSection
+                data={this.props[fieldNames.sections.infographicsSection]}
+                moreURI={`categories/${categoryConst.pathSegments.infographic}`}
+              />
+            </Background>
+          </Section>
```
babygoat, line 312: Q: what if we use the inner section directly without `AnchorWrapper`? Will we have any warning here?

src/components/side-bar/with-anchors.js
```diff
@@ -0,0 +1,113 @@
...
+ * }, {
+ *  id: 'section-3',
+ *  label: 'Section Three'
+ * }]
+ * ```
+ *
+ * @param {Object} DecoratedComponent A React component wants to have side bar functionality
+ * @returns {Object} A React component with anchors props
+ */
+function withAnchors(DecoratedComponent) {
```
babygoat, line 82: These codes are similar to these in @twreporter-npm-packages/react-components/src/side-bar. Why do we choose to implement it independently rather than in the SideBarHOC?

src/containers/Home.js
taylrj: If we use inner section directly, the sections should be configured as :

```<LatestSection 
        data={this.props[fieldNames.sections.latestSection]}
        anchor-id="latest"
        anchor-label="最新消息"
      />
```

I think the way using `AnchorWrapper` would be easier to understand and more declarative.

src/components/side-bar/with-anchors.js
taylrj: Firstly, I wanted to separate functions `collecting anchors` from `generating side bar`.

But I'm considering this suggestion to merge `withAnchors` into `SideBarHOC` since if it is a good practice to declare all the anchors via the way of this change, we could make `SideBarHOC` to adopt such usage by default.

taylrj:
Updated.

I moved changes above into https://github.com/twreporter/twreporter-npm-packages/pull/186, please help to review as well.

taylrj:
Updated.

Move changes above into https://github.com/twreporter/twreporter-npm-packages/pull/186.

Also, use `showAnchor` prop to indicate whether to show anchor on side bar instead of using empty 'label'.

src/components/side-bar/base-components.js
```diff
@@ -73,9 +73,9 @@ class Anchors extends React.PureComponent {
     data.forEach(anchorObj => {
       const id = _.get(anchorObj, 'id', '')
       const label = _.get(anchorObj, 'label', '')
+      const show = _.get(anchorObj, 'show', true)
 
-      // id and label are not empty string
-      if (id && label) {
+      if (show && id && label) {
```
babygoat, line 78: If the `show` property is used for the indicator to display,  why not `if (show)`?

taylrj: Sure, I think `show` is enough here.

Updated.

src/containers/Home.js
```diff
@@ -295,39 +255,61 @@ class Homepage extends React.PureComponent {
...
-          />
-          <NewsLetterSection />
-          <Background backgroundColor={moduleBackgounds.category}>
-            <CategorySection data={this.props.categories} />
-          </Background>
-          <Background backgroundColor={moduleBackgounds.topic}>
-            <TopicsSection
-              data={this.props[fieldNames.sections.topicsSection]}
+        <SideBar ref={this._sidebar}>
+          <Section anchorId="latest" anchorLabel="" showAnchor={false}>
```
babygoat, line 259: minor. see here for [reference](https://github.com/twreporter/twreporter-npm-packages/pull/186#discussion_r630725234)

taylrj: Updated.

Issue #1846, opened 2021-05-05
Desktop sidebar overlaps on the footer


achinchen:
## Bug report
Browsing website and found sidebar layout issue.

### Description 
Found desktop sidebar overlaps on the footer.

### Steps to reproduce
Using screen width wider than 768px, and scroll page to bottom.

### Actual behavior
Sidebar overlays on the footer.

### Expected behavior
Sidebar doesn't overlay on the footer.

### Screenshots
![Untitled](https://user-images.githubusercontent.com/40432044/117092112-ee9b0f00-ad8f-11eb-87da-2f856f228110.png)

### Possible solution
Update sidebar style.
If the behavior I referred above is accepted, I'm happy to make PR to update it.

taylrj:
Thanks for the report.

The expected behavior referred above is acceptable.
Feel free to make a pull request or comments below.

Thank you~

achinchen:
@taylrj I opened the PR to fix it issue, please review it.

PR #1850, opened 2021-05-09
fix(homepage): prevent sidebar from overlapping footer

achinchen:
## Changes 

- Remove legacy commented styles.
- Replace unused div with React.Fragment.
- Adjust vertical center style in `index-page-side-bar`.
- Remove `overflow: hidden` from `container`.

## Description

According to best practice `graceful degradation`, using position `sticky`(for modern browser) + `fixed`(for legacy browser) to fix this styling issue.

## Screenshot
image
image

## Related Issue 
Resolves: #1846

### DIFF
--- a/src/components/side-bar/index-page-side-bar.js
+++ b/src/components/side-bar/index-page-side-bar.js
@@ -1,12 +1,10 @@
 import PropTypes from 'prop-types'
-import React from 'react'
+import React, { Fragment } from 'react'
 import baseComponents from './base-components'
 import colors from '../../constants/colors'
 import mq from '../../utils/media-query'
 import styled from 'styled-components'
 
-// writing-mode: vertical-rl;
-// letter-spacing: 2px;
 const StyledAnchor = styled(baseComponents.StyledAnchor)`
   margin-bottom: 18px;
   color: ${props => (props.highlight ? 'white' : `${colors.primaryColor}`)};
@@ -22,11 +20,17 @@ class Anchors extends baseComponents.Anchors {
 
 const SideBarContainer = styled.div`
   position: fixed;
-  top: 50%;
+  @supports (position: sticky) {
+    position: sticky;
+  }
+
+  float: right;
+  right: 16px;
+  top: 50vh;
   z-index: 100;
   transform: translateY(-50%);
   color: ${colors.primaryColor};
-  right: 16px;
+  will-change: transform;
   ${mq.tabletOnly`
     right: 3px;
   `}
@@ -40,7 +44,7 @@ class HomePageSideBar extends React.PureComponent {
     // currentAnchorId and handleClickAnchor are passed from `SideBarHOC`
     const { anchors, children, currentAnchorId, handleClickAnchor } = this.props
     return (
-      <div>
+      <Fragment>
         <SideBarContainer>
           <Anchors
             data={anchors}
@@ -49,7 +53,7 @@ class HomePageSideBar extends React.PureComponent {
           />
         </SideBarContainer>
         {children}
-      </div>
+      </Fragment>
     )
   }
 }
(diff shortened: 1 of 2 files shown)


src/components/side-bar/index-page-side-bar.js
```diff
@@ -22,9 +20,7 @@ class Anchors extends baseComponents.Anchors {
 
 const SideBarContainer = styled.div`
   position: fixed;
-  top: 50%;
   z-index: 100;
-  transform: translateY(-50%);
```
taylrj, line 27: It seems the issue still exists even if we remove vertical center style in some device proportions (e.g. 1280px x 698px).

I came up with another approach:

```js
const SideBarContainer = styled.div`
  /* The fallback goes first */
  position: fixed;
  /* This 'enhancement' is ignored if not understood (e.g. IE 11), */
  /* but overrides the previous value if the browser supports it. */
  position: sticky;
  float: right;
  top: 50vh;
  /*...etc */
`
```

On one hand, we can prevent sidebar from overlapping footer when it supports `position: sticky`. On the other hand, keep the behavior as it is when `position: sticky` is not supported.

## Screenshot
image 

 @achinchen Does this idea also make sense to you ?

taylrj: Further to my previous comment, we still need to remove [this line](https://github.com/twreporter/twreporter-react/blob/3877f99b04a9a7e782dee177effb0985b3acb11a/src/containers/Home.js#L88) with this approach.

achinchen: @taylrj 
Thanks for advice, I made some changes to prevent `float` and be more self-documenting.
Please review the changes, thanks!

taylrj: Thanks for your effort to make it more self-documenting.

However, there is concern that preventing `float` and adding `display: inline-block` result in an inline-level block side bar container which still takes up a whole block.

### screenshot 
image

achinchen: @taylrj I try several way to fix the issue, but it looks like `add float` would make thing simple.
Moreover, I tried to tidy up the styles and optimize performance for sticky, I reserved `will-change: transform`
There's the [REF](https://developer.mozilla.org/en-US/docs/Web/CSS/position#performance_accessibility)

taylrj: Thanks for the reference.

🙋‍♂️  Last question, if we add `will-change: transform` to the positioned elements to render the element in its [own layer](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) to improve repaint speed. Why do we need to add `right` as the property value of `will-change` as well ?

achinchen: @taylrj 
Thank you for patiently reviewing.

Due to resize window causes right value to change, though it's very minor. In my opinion, I'd like to add it for optimization.
If there's any concern, let me know, even though removing it isn't a big thing for me.

achinchen: BTW, it's glad to hear you mentioned `stacking context`, that's awesome topic for frontend dev.✨

taylrj: Thanks for the explanation. It's clear.

After a second thought, I think we can follow the suggestion from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change):

> Important: will-change is intended to be used as a last resort, in order to try to deal with existing performance problems. It should not be used to anticipate performance problems.

In view of the suggestion above, `will-change: transform` looks good to me since it helps browser to manage repaints at 60 fps. It is an existing performance issue.
Follow the same suggestion, I prefer to remove `right` from values of `will-change` due to resize window causes value change is a minor case as mentioned in previous comment and it's more like an anticipate performance problem to me.

taylrj (approved):
Thank you @achinchen !!

LGTM

PR #1865, opened 2021-05-26
fix: fix slideshow effect for mobile devices on index page

taylrj:
## Changes
- add `overflow: hidden` to index page container to fix slideshow effect

## Description
On [staging.twreporter.org](https://staging.twreporter.org), slideshow effect does not work as expected on mobile devices.
The misbehavior can only be reproduced on real mobile devices, not the browser emulator.

### screenshot
IMG_6E1F7044F241-1

### DIFF
--- a/src/containers/Home.js
+++ b/src/containers/Home.js
@@ -8,6 +8,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import categoryConst from '../constants/category'
 import loggerFactory from '../logger'
+import mq from '../utils/media-query'
 import qs from 'qs'
 import sideBarFactory from '../components/side-bar/side-bar-factory'
 import siteMeta from '../constants/site-meta'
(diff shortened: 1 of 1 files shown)


src/containers/Home.js
```diff
@@ -85,6 +86,9 @@ const Container = styled.div`
   width: 100%;
   margin: 0 auto;
   background-color: white;
+  ${mq.mobileOnly`
+    overflow: hidden;
```
babygoat, line 90: What if we make it `overflow-x: hidden`?

taylrj: In my experiment, it also breaks `position: sticky;` on desktop. (tested on Chrome, Firefox)

babygoat:
LGTM

src/containers/Home.js
babygoat: Ya, my fault. Since overflow/overflow-x/overflow-y: hidden will always try to create a new scroll container.  It might not be possible to make the sticky attribute stick to document.

PR #1876, opened 2021-06-03
refactor: remove the default email of authors

babygoat:
This patch removes the default email(contact@twreporter.org) of authors.
Address the inquiry from the editors.

issue: twreporter-36


taylrj:
LGTM

PR #1877, opened 2021-06-03
fix: eslint warnings

babygoat:
This patch fixes several eslint warnings and adds the generated
sw.js to ignore error.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ lib
 public
 static
 out
+sw.js
--- a/service-worker/service-worker-generator.js
+++ b/service-worker/service-worker-generator.js
@@ -47,7 +47,7 @@ function generateRuntimeCaching(runtimeCaching) {
 }
 
 const releaseBranch = config.releaseBranch
-const origin = url.parse(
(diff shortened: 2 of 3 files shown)


babygoat:
@taylrj  Only react/no-find-dom-node warnings are not resolved. I'm curious why u prefer to enforce the rule explicitly. Do u have any plans to migrate the API to react refs?

taylrj:
LGTM.

Yes, there's a plan to migrate the API since `findDOMNode` is going to be deprecated.
The alternative API usage needs some investigation, we don't have to include the change in this PR.

babygoat:
The reason why I do the eslint fix is that I would like to enable the linter on CI but the warnings will break it. AFAIK, we need to put the refs on the child elements and use it for offset and height calculation, right?

taylrj:
> AFAIK, we need to put the refs on the child elements and use it for offset and height calculation, right?

In theory, yes.

But I haven't figure out the way doing so and keep current behavior at the same time.
It might need more time to investigate how to refactor away use of findDOMNode.

PR #1878, opened 2021-06-09
feat: manually send gtm.load event for SPA

babygoat:
This patch sends the `gtm.load` event to reset gtm scroll depth as sonn
as the window is ready.


babygoat:
I will deploy the patch for the test on staging first. Please comment here if you have any suggestions and I will address them in the following patches.

PR #1882, opened 2021-06-10
fix: send gtm.load event when the content is ready

babygoat:
This patch sends the gtm.load event when the content is ready.


src/containers/Home.js
```diff
@@ -205,6 +200,17 @@ class Homepage extends React.PureComponent {
     })
   }
 
+  componentDidUpdate(prevProps) {
+    if (!prevProps.isContentReady && this.props.isContentReady) {
```
babygoat, line 204: @taylrj  let me know if you have any concerns about introducing the new prop.

babygoat:
I will merge the patch first for testing on staging.

taylrj:
LGTM

Issue #1871, opened 2021-06-01
Broken search UI on topic landing page in Firefox


taylrj:
### Description 
Found broken search UI in Firefox

### Steps to reproduce

1. open topic landing page (e.g. https://www.twreporter.org/topics/rent-house-in-taiwan)
2. click search icon (🔎)
3. leave the search bar empty and click elsewhere

### Actual behavior
The search icon doesn't display.

### Expected behavior
The search icon displays again.

### Screenshots
![image](https://user-images.githubusercontent.com/6375655/120274823-fb554900-c2e2-11eb-8664-23ee0c68df00.png)
![image](https://user-images.githubusercontent.com/6375655/120274861-0ad49200-c2e3-11eb-8d8d-217bd12c3cfe.png)

PR #1898, opened 2021-07-09
fix(search-box): give input a specific width to avoid bug in firefox

wtlin1228:
## Changes

- Add width for input in the search box

## Description

Input's width won't be limited by the parent's width In Firefox. So make the input 200px can fix the bug in Firefox.

## Related Issue

Resolves: #1871

### DIFF
--- a/src/components/topic/search-box.js
+++ b/src/components/topic/search-box.js
@@ -46,6 +46,7 @@ const Input = styled.input`
   outline: none;
   border: none;
   height: 100%;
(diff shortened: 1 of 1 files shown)


taylrj (approved):
LGTM

Thanks for the fix, @wtlin1228 !

taylrj:
https://github.com/twreporter/twreporter-react/pull/1910 landed 🚀

PR #1936, opened 2021-09-23
fix: call to action banner can not be clicked

taylrj:
Address [TWREPORTER-140](https://twreporter-org.atlassian.net/browse/TWREPORTER-140?atlOrigin=eyJpIjoiNDczOGVjOTgxMGEyNDQwZTk4NTMyMTQ4ZTg4NzY0NDMiLCJwIjoiaiJ9)

This patch removes the wrapper of header in theme transparent,
it fixes a bug where the CTA banner can not be clicked.

### DIFF
--- a/src/containers/app-shell.js
+++ b/src/containers/app-shell.js
@@ -2,7 +2,6 @@ import { connect } from 'react-redux'
 import ErrorBoundary from '../components/ErrorBoundary'
 import Footer from '@twreporter/react-components/lib/footer'
 import Header from '@twreporter/universal-header/lib/containers/header'
-import mq from '@twreporter/core/lib/utils/media-query'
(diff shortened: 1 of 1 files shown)


Aylie-Chou:
Do you know why we need the wrapper if transparent theme has already been a  regular theme ?

taylrj:
I think we need the wrapper because it helps current header to be in position 'absolute' instead of 'relative', which leads to a styling issue according to the screenshot.

image

taylrj:
> I think we need the wrapper because it helps current header to be in position 'absolute' instead of 'relative', which leads to a styling issue according to the screenshot.

The issue seems to exist in header 2.0, so I am closing this PR.

Will reopen when it is resolved.

PR #1937, opened 2021-09-27
feat: update topic landing page header to universal-header

Aylie-Chou:
This patch adapt universal header to topic landing page.


taylrj:
LGTM

PR #1945, opened 2021-09-29
fix: update comments to single lines

taylrj:
This patch updates comments in web-push component in single-lines
instead of multi-line since styled-compnent does not support.


Aylie-Chou (approved):
LGTM

PR #1949, opened 2021-09-29
fix: fix header/web-push z-index & transparent warpper style

Aylie-Chou:
- z-index setting
  - web-push: 1001
  - header: 1000
  - other component: 999 or below


taylrj:
LGTM

PR #2021, opened 2021-12-09
feat: add user id datalayer

Aylie-Chou:
This patch push user id datalayer for gtm analysis & separate gtm data to different environments.

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -11,9 +11,29 @@ import hashLinkScroll from './utils/hash-link-scroll'
 import loggerFactory from './logger'
 import releaseBranchConsts from '@twreporter/core/lib/constants/release-branch'
 import twreporterRedux from '@twreporter/redux'
+// lodash
+import get from 'lodash/get'
+const _ = {
+  get,
+}
 
 const logger = loggerFactory.getLogger()
 const releaseBranch = globalEnv.releaseBranch
+const tagManagerArgs = {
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -11,9 +11,29 @@ import hashLinkScroll from './utils/hash-link-scroll'
...
+const _ = {
+  get,
+}
 
 const logger = loggerFactory.getLogger()
 const releaseBranch = globalEnv.releaseBranch
+const tagManagerArgs = {
+  master: {
+    gtmId: 'GTM-PRMXBBN',
+    auth: '2pJC7GotZqWa7HtmIgSFIg',
```
taylrj, line 25: Could you share the idea about why do we need `auth`, `preview` and why do they are different corresponds to different environments (master, staging) ?

Aylie-Chou: The official doc about environments can be found [here](https://support.google.com/tagmanager/answer/6311518).

The main reason we need to include environments is to publish testing gtm version to testing environment, thus we can show test data on GA matrix.
I separate master & staging for corresponding to releaseBranch in codebase. They can be combined into a single testing environment.

How do you consider about it ?

taylrj: I second this idea.

We can separate master & staging as different environments.

taylrj:
LGTM

PR #2048, opened 2022-02-25
fix: add gtm preview environment config

Aylie-Chou:
this patch fix `Uncaught TypeError: Cannot read properties of undefined` in keyston-preview.


taylrj:
LGTM

PR #2051, opened 2022-02-25
fix: fix eslint error

taylrj:
This patch fixes lint error as follows:
- 'children' is missing in props validation
- Expected error to be handled
- 'mq' is defined but never used


duidae (approved):
checked with `make lint` & errors are gone. 👍

PR #2123, opened 2022-05-10
fix: change web font loader to address typekit embedded code issue

taylrj:
Address [TWREPORTER-318](https://twreporter-org.atlassian.net/browse/TWREPORTER-318)

This patch adopts
[webfontloader](https://github.com/typekit/webfontloader) to load
typekit web font to address an issue in Safari, especially mobile
Safari, where white spaces keep occupying screen when user touch
moving.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -94,17 +94,20 @@ export default class Html extends PureComponent {
           {_.map(scripts, (script, key) => (
             <script src={script} key={'scripts' + key} charSet="UTF-8" />
           ))}
-
           {/* <!-- Load typekit fonts for twreporter.org domain--> */}
           <script
             dangerouslySetInnerHTML={{
-              __html: `(function(d) {
-                var config = {
-                kitId: 'vlk1qbe',
-                scriptTimeout: 3000,
-                async: true
-              },h=d.documentElement,t=setTimeout(function(){h.className=h.className.replace(/\bwf-loading\b/g,"")+" wf-inactive";},config.scriptTimeout),tk=d.createElement("script"),f=false,s=d.getElementsByTagName("script")[0],a;h.className+=" wf-loading";tk.src='https://use.typekit.net/'+config.kitId+'.js';tk.async=true;tk.onload=tk.onreadystatechange=function(){a=this.readyState;if(f||a&&a!="complete"&&a!="loaded")return;f=true;clearTimeout(t);try{Typekit.load(config)}catch(e){}};s.parentNode.insertBefore(tk,s)
-            })(document);`,
+              __html: `
+                WebFontConfig = {
+                  typekit: { id: 'vlk1qbe' }
+                };
+                (function(d) {
+                   var wf = d.createElement('script'), s = d.scripts[0];
+                   wf.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.26/webfont.js';
+                   wf.async = true;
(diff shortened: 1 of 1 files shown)


src/html.js
```diff
@@ -94,17 +94,20 @@ export default class Html extends PureComponent {
...
           {/* <!-- Load typekit fonts for twreporter.org domain--> */}
           <script
             dangerouslySetInnerHTML={{
-              __html: `(function(d) {
-                var config = {
-                kitId: 'vlk1qbe',
-                scriptTimeout: 3000,
-                async: true
-              },h=d.documentElement,t=setTimeout(function(){h.className=h.className.replace(/\bwf-loading\b/g,"")+" wf-inactive";},config.scriptTimeout),tk=d.createElement("script"),f=false,s=d.getElementsByTagName("script")[0],a;h.className+=" wf-loading";tk.src='https://use.typekit.net/'+config.kitId+'.js';tk.async=true;tk.onload=tk.onreadystatechange=function(){a=this.readyState;if(f||a&&a!="complete"&&a!="loaded")return;f=true;clearTimeout(t);try{Typekit.load(config)}catch(e){}};s.parentNode.insertBefore(tk,s)
-            })(document);`,
```
Aylie-Chou, line 107: Question: What makes old method probleming since it seems to load fonts asynchronously too ?

src/html.js
```diff
@@ -94,17 +94,20 @@ export default class Html extends PureComponent {
...
+              __html: `
+                WebFontConfig = {
+                  typekit: { id: 'vlk1qbe' }
+                };
+                (function(d) {
+                   var wf = d.createElement('script'), s = d.scripts[0];
+                   wf.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.26/webfont.js';
+                   wf.async = true;
+                   s.parentNode.insertBefore(wf, s);
+                })(document);
```
Aylie-Chou, line 109: Question: Would we need to deal with `FOUT` problem while using new method ?
(old method add `FOUT` by adding wf-* class ([ref](https://blog.typekit.com/2011/05/25/loading-typekit-fonts-asynchronously/)))

taylrj:
Thanks for all the questions above, they are helpful to me doing more research about this issue.

Indeed, I did a little digging and experiment, the root cause has nothing to do with the web font loader.

Hence, I'm going to close this PR.

Currently, I believe this issue is caused by the font 'source-han-sans-traditional'.

After `document.fonts` add such font, the bizarre behavior occurs in Mobile Safari.

It can be reproduced on localhost.

image

src/html.js
taylrj: Yes, it also loads font asynchronously.

But based on my observation, the bizarre behavior took place after the font is loaded. 

IMO, the issue is about what's the adobe font script does after the font is loaded, which is still not clear why it leads to such bizarre behavior.

For me, it is hard to trace the code `https://use.typekit.net/vlk1qbe.js` since it is minimized and uglified.

I proposed a similar solution in #2148 to address this issue, which uses web font loader and google font instead to mitigate the issue.

src/html.js
taylrj: Yes, the method it uses to address FOUT problem is the same with the old method.

That is,

Add 'wf-loading' class to html when the font is loading, remove it and add class 'wf-xxxxx-active' when the fonts loaded successfully.

PR #2130, opened 2022-05-24
refactor: use inline style changing widith of reading progress

taylrj:
Address [TWREPORTER-318](https://twreporter-org.atlassian.net/browse/TWREPORTER-318)

This patch uses the `attrs` method, together with a style object
changes the in-line style for frequently changing width of reading
progresss instead of generating unique classes whenever its style
changes since generating CSS classes isn't a lightweight operation.

Please note that this change aims to mitigate the 'blank page' issue in
iOS Safari/Chrome/Firefox where renders white spaces on part of screen.

### DIFF
--- a/src/components/article/reading-progress.js
+++ b/src/components/article/reading-progress.js
@@ -2,16 +2,17 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
 
-const Bar = styled.div`
+const Bar = styled.div.attrs(({ percent }) => ({
+  style: {
+    width: percent + '%',
+  },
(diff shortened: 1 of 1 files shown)


duidae:
@taylrj nice catch, and

1. Would you mind explaining a bit more about this fix, such as how to spot the issue/how this fix would work?
2. This part of code looks still implicit after the bug is fixed, could we add some comments as a warning, in case someone else accidentally changes the code and breaks it again?

taylrj:
Close due to spec change. The reading progress component should be removed. Thanks for comments, @duidae .

PR #2131, opened 2022-05-25
refactor: remove reading progress due to spec change

taylrj:
This patch removes reading progress component due to spec change of
article page.


duidae (changes requested):
looks good, and could we also clean up the related event listener part of `handleScroll` in `componentWillUnmount`, thanks!

https://github.com/twreporter/twreporter-react/blob/65c9523024bd51b41080b291453ecebbc14cf83a/src/containers/Article.js#L80

taylrj:
Oops, thanks.

Updated.

duidae (approved):

PR #2192, opened 2022-07-18
fix: remove smoothscroll lib

duidae:
P.S. this PR will add corresponding updates of npm packages later.


src/components/about-us/index.js
```diff
@@ -46,7 +45,10 @@ export class AboutUs extends PureComponent {
     this.sectionOffset = []
   }
   _handleClickAnchor = anchorIdx => {
-    return smoothScroll(this.sectionOffset[anchorIdx])
+    window.scroll({
+      top: this.sectionOffset[anchorIdx],
+      behavior: 'smooth',
```
taylrj, line 50: It seems to me that [Safari does not have support for the smooth scroll behavior](https://developer.mozilla.org/en-US/docs/Web/API/Window/scroll#browser_compatibility) ... 🥲

duidae: oh no...:cry

taylrj: https://github.com/iamdustan/smoothscroll

polyfill ?

duidae: replaced with our custom smooth scroll.

PR #2230, opened 2022-09-06
refactor: async load google optimize script

taylrj:
Address [TWREPORTER-423](https://twreporter-org.atlassian.net/browse/TWREPORTER-423)

image

This change async loads google optimize script to eliminate
render-blocking resources.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -27,7 +27,10 @@ export default class Html extends PureComponent {
     return (
       <html lang="zh-TW">
         <head>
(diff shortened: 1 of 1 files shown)


taylrj:
Note: According to the [official suggestion](https://support.google.com/optimize/answer/7513085), the anti-flicker snippet will be installed only when needed.

duidae (approved):
LGTM

PR #2242, opened 2022-09-15
fix: photography page data loader

taylrj:
This patch fixes data loader for photography page where it can not get posts on server side rendering.


Aylie-Chou (approved):
LGTM 🙏

PR #2246, opened 2022-09-19
fix: apply new category set in index-page & photography page

Aylie-Chou:
This patch add apply category set on index page & photography page with feature toggle.
And correspond with twreporter-npm-packages [#310](https://github.com/twreporter/twreporter-npm-packages/pull/310)


duidae (approved):
LGTM

PR #2266, opened 2022-10-16
refactor: adopt adobe font as the third-party web fonts

taylrj:
Address [Reduce unused CSS: google font CSS link](https://app.asana.com/0/0/1202942985879129/f)

Note that https://github.com/twreporter/twreporter-npm-packages/pull/314 is needed by this change.

This patch regresses to adopt adobe font as the third-party web fonts, here is the reasons:

- In previous change, we used google font to replace adobe font to address the render blocking issue which leads to white page appears while scrolling in iOS browsers.
- After a while, we found that the cause of that issue was the subsetting of a specific font 'Noto Sans TC' and we decided to self-host that font on gcs as a whole (without subsetting).
- So we can see adobe font service has nothing to do with this issue.
- This change aims to improve the performance of the website suggested by PageSpeed Insight, which were:
  - Reduce unused CSS because it is render blocking resource: Google fonts
  - Reduce the impact of third-party code which may block main thread: webfont.js
- Therefore, we regress to adobe font which does not uses CSS but a async script instead and does not need the webfont.js which is a third-party font loader.

## Reference
- Reduce unused CSS
image

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -110,39 +110,19 @@ export default class Html extends PureComponent {
           {_.map(scripts, (script, key) => (
             <script src={script} key={'scripts' + key} charSet="UTF-8" />
           ))}
-          {/* <!-- Load google fonts --> */}
           <script
             dangerouslySetInnerHTML={{
-              __html: `
-              (function(document) {
-                var gf = document.createElement("link"),
-                gfApi = document.createElement("link");
-                gf.rel="dns-prefetch";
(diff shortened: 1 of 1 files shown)


duidae (approved):

taylrj:
Upgrade npm packages in https://github.com/twreporter/twreporter-react/pull/2287, which is the change in https://github.com/twreporter/twreporter-npm-packages/pull/314.

PR #2300, opened 2022-11-18
fix: adjust about us department

Aylie-Chou:
- engineering -> digital product
- remove product manager
- add data journalism


duidae (approved):
LGTM!

PR #2331, opened 2022-12-23
fix: install gke auth plugin

duidae:


duidae:
https://github.com/twreporter/twreporter-react/issues/2337

PR #2343, opened 2022-12-29
feat: add latest page

Aylie-Chou:
This patch add latest page.
To test it, please correspond to [`twreporter-npm-packages` PR #345](https://github.com/twreporter/twreporter-npm-packages/pull/345)

### DIFF
--- a/src/containers/Category.js
+++ b/src/containers/Category.js
@@ -246,8 +246,7 @@ function titleTabProp(categoryPath, subcategoryPath) {
       subcategoryKey !== 'all'
         ? `/categories/${categoryPath}/${path}`
         : `/categories/${categoryPath}`
-    const isExternal = false
-    return { id, label, link, path, isExternal }
+    return { id, label, link, path, isExternal: false }
   })
   const activeTabIndex = subcategoryPath
     ? _.findIndex(subcategoryList, ['path', subcategoryPath])
--- a/src/containers/Latest.js
+++ b/src/containers/Latest.js
@@ -18,27 +18,24 @@ import twreporterRedux from '@twreporter/redux'
 import mq from '@twreporter/core/lib/utils/media-query'
 import { List } from '@twreporter/react-components/lib/listing-page'
 import { TitleTab } from '@twreporter/react-components/lib/title-bar'
-import {
-  CATEGORY_ID,
-  CATEGORY_LABEL,
-  SUBCATEGORY_ID,
-  SUBCATEGORY_LABEL,
-  CATEGORY_SET,
-} from '@twreporter/core/lib/constants/category-set'
 // lodash
 import forEach from 'lodash/forEach'
 import get from 'lodash/get'
 import map from 'lodash/map'
 import findIndex from 'lodash/findIndex'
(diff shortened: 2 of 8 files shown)


src/containers/Category.js
```diff
@@ -232,6 +232,30 @@ function postsProp(state, listId, page) {
...
+  const subcategoryList = _.map(CATEGORY_SET[categoryPath], subcategoryKey => {
+    const path = subcategoryKey
+    const label = SUBCATEGORY_LABEL[subcategoryKey]
+    const id = SUBCATEGORY_ID[subcategoryKey]
+    const link =
+      subcategoryKey !== 'all'
+        ? `/categories/${categoryPath}/${path}`
+        : `/categories/${categoryPath}`
+    const isExternal = false
+    return { id, label, link, path, isExternal }
```
duidae, line 250: `isExternal` is a constant & we only use once, so probably we could just replace this variable with `false` & use comment to show it's a boolean meaning `isExternal`

src/containers/Latest.js
```diff
@@ -0,0 +1,348 @@
...
+    fetchPostsByTagListId,
+    fetchLatestPosts,
+  ])
+
+  // Error handling
+  if (error) {
+    return <SystemError error={error} />
+  }
+
+  const title = 'latest' + siteMeta.name.separator + siteMeta.name.full
```
duidae, line 120: 'latest' showed up twice(line 299), so we could replace them with constant variable to reduce duplications

duidae (changes requested):

src/containers/Latest.js
```diff
@@ -0,0 +1,348 @@
...
+
+const TitleTabContainer = styled.div`
+  margin: 64px 0;
+
+  ${mq.tabletOnly`
+    margin: 32px 0;
+  `}
+  ${mq.mobileOnly`
+    margin: 24px 0;
+    width: 86.66666666666667%;
```
duidae, line 67: not so important, but curious about this setting, do we really need so many digits? 😂 what about just use 86.67%

src/containers/Category.js
Aylie-Chou: updated ([ac2670c](https://github.com/twreporter/twreporter-react/pull/2343/commits/ac2670c5c5e4145142df2d4b7a2c6fbbb804121c))

src/containers/Latest.js
Aylie-Chou: updated ([ac2670c](https://github.com/twreporter/twreporter-react/pull/2343/commits/ac2670c5c5e4145142df2d4b7a2c6fbbb804121c))
`latest` in line 299 means its redux action list id, thus I export new `LATES_LIST_ID` for it (referr to twreporter-npm-packages [#352](https://github.com/twreporter/twreporter-npm-packages/pull/352))

src/containers/Latest.js
Aylie-Chou: Good point 😆
It's just the copy code from previous...
I think we could manage it when the layout need to change.

duidae (approved):

PR #2355, opened 2023-01-09
fix: modify circleci config for new context

Aylie-Chou:
- use new context twreporter-gcp
- rename GOOGLE_PREVIEW_CLUSTER_NAME to K8S_PREVIEW_CLUSTER_NAME in new context


howar31 (approved):
LGTM

PR #2376, opened 2023-02-17
fix: sync node version with npm packages

duidae:
sync with `twreporter-npm-packages`


.circleci/config.yml
```diff
@@ -6,8 +6,7 @@ version: 2
 jobs:
   build:
     docker:
-      # specify the version you desire here
-      - image: circleci/node:10.15
+      - image: cimg/node:16.14.2
```
Aylie-Chou, line 9: The warning from linter would also need to be resolved before upgrading to new version.
`warning  'url.parse' was deprecated since v11.0.0. Use 'url.URL' constructor instead`

Aylie-Chou (changes requested):

.circleci/config.yml
duidae: fixed, thanks. btw how do you get the warning?

Aylie-Chou: `npm run lint`

Aylie-Chou (approved):
LGTM

Issue #2430, opened 2023-03-18
macos 13.2.1 fails make dev


sharils:
macos 13.2.1 `make dev` log:

```
❯ make dev
delete auto generated files, including processes.json, sw.js, dist/, webpack-assets.json and react-loadabel.json

Setup development environment.
Development environment will contains three different servers.
One will be application server, hosted on 3000 port.
Another will be webpack dev server, hosted on 5000 port.
The other will be mocked api server, hosted on 8080 port.
[1]  [+] start dev server by nodemon src/server.js
[1]
[1] NODE_ENV=development RELEASE_BRANCH= RENDER_ENV=server HOST= node_modules/.bin/nodemon src/server.js --exec node_modules/.bin/babel-node
[2]  [+] start testing server by babel-node src/testing-server.js
[2]
[0]  [+] start webpack dev server by node webpack-dev-server.js
[0]
[0] NODE_ENV=development RELEASE_BRANCH= DEV_HOST= node webpack-dev-server.js
[1] [nodemon] 1.18.10
[1] [nodemon] to restart at any time, enter `rs`
[1] [nodemon] watching: *.*
[1] [nodemon] starting `node_modules/.bin/babel-node src/server.js`
[0] (node:11425) [DEP0111] DeprecationWarning: Access to process.binding('http_parser') is deprecated.
[0] (Use `node --trace-deprecation ...` to show where the warning was created)
[0] node:events:490
[0]       throw er; // Unhandled 'error' event
[0]       ^
[0]
[0] Error: listen EADDRINUSE: address already in use 0.0.0.0:5000
[0]     at Server.setupListenHandle [as _listen2] (node:net:1740:16)
[0]     at listenInCluster (node:net:1788:12)
[0]     at doListen (node:net:1937:7)
[0]     at process.processTicksAndRejections (node:internal/process/task_queues:83:21)
[0] Emitted 'error' event on Server instance at:
[0]     at emitErrorNT (node:net:1767:8)
[0]     at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
[0]   code: 'EADDRINUSE',
[0]   errno: -48,
[0]   syscall: 'listen',
[0]   address: '0.0.0.0',
[0]   port: 5000
[0] }
[0]
[0] Node.js v19.6.0
[0] make[1]: *** [start-webpack-dev-server] Error 1
[0] /Applications/Xcode.app/Contents/Developer/usr/bin/make start-webpack-dev-server exited with code 2
[...]

PR #2431, opened 2023-03-18
Change webpack dev server port #2430

sharils:


sharils:
@Aylie-Chou Would you mind reviewing this instead? Thanks.

sharils:
Hi @Aylie-Chou I understand that your team is busy, but would you be possible to leave a message about this? So we can move this forward? Thanks.

Aylie-Chou:
hi @sharils , sorry for the late reply, I was on a vacation 😂
This configs change looks good to me on mac environment !
But I think we woul need to survey wether this change also works on other environment (including our deploy environment).
I would discuss to our team to isee f we can cover it in next sprint. I would back to update this PR after that.

PR #2433, opened 2023-03-25
fix: add font smoothing

duidae:
# Related Issue

https://app.asana.com/0/1203699264568808/1203983432530619/f

# Root Cause

# Solution


Aylie-Chou (approved):
LGTM

PR #2455, opened 2023-05-07
fix: add webpack loader for css & font files

Aylie-Chou:
# Issue
- add webpack loader for css files to support material symbols (font type)
- material symbol packages use `@font-face` syntax, need to import in root file
  - move it from @twreporter/react-components to `src/client.js`

# Notice
`membership-fe` should add same import to use material symbols

# Dependency
twreporter-npm-packages [#385](https://github.com/twreporter/twreporter-npm-packages/pull/385)


liruchen32 (approved):
ＬＧＴＭ

PR #2495, opened 2023-05-29
Update package.json

liruchen32:
# Issue
update `package.json` for `twreporter-npm-packages`

# Notice

# Dependency


Aylie-Chou (changes requested):
The lock file `yarn.lock` would also need to be updated

PR #2501, opened 2023-05-30
fix: fix new URL type problem

liruchen32:
# Issue
`Server error was caught by Express default error handler. Invalid URL: / TypeError [ERR_INVALID_URL]: Invalid URL: /`
# Notice

# Dependency


Aylie-Chou (approved):
LGTM.
Could you help to check if other `new URL` part has same issue or not ?

PR #2511, opened 2023-06-06
Feature/email subscription page

liruchen32:
# Issue
[[會員制] 電子報訂閱變更頁](https://app.asana.com/0/1203699264568808/1204692921007274/f)
# Notice

# Dependency
[npm-packages#399](https://github.com/twreporter/twreporter-npm-packages/pull/399)


Aylie-Chou (changes requested):
Mostly looking good 👍
Current bookmark url would use `/bookmarks/:pageNumber?` defined in `constants/routes`. It's good to use the defined value than plain text, but could we hide the parameter definition ?
ex. use `/bookmarks` or `/bookmarks/1`

Aylie-Chou (changes requested):
current path in `constant/route.js` is for page routing, thus it may includ some parameters definitions, and convert to props value for different pages.
When we assign the url path, the value would also need to be assigned (or use default value).
There're two possible way to address it :
- store another field in `routes.bookmarkListPage`
- slice the path

Aylie-Chou (approved):
LGTM

PR #2519, opened 2023-06-14
fix: import winston module with require

Aylie-Chou:
# Issue
fix import warning ([ref](https://cloud.google.com/nodejs/docs/reference/logging-winston/latest))

螢幕截圖 2023-06-14 下午3 26 34


liruchen32 (approved):
LGTM

PR #2530, opened 2023-06-20
fix: apply `babel-loader` on react-article-components

Aylie-Chou:
# Issue
[20230531 文章頁 reload error](https://app.asana.com/0/0/1204754585032337/f)

# Reason
`react-article-components` has some async functions need to convert to es5 syntax with babel.


liruchen32 (approved):
LGTM

PR #2534, opened 2023-06-21
feat: get user data from state

liruchen32:
# Issue
[[會員制] 會員專區的會員身份 API 串接](https://app.asana.com/0/1203699264568808/1204362012792382/f)

# Notice

# Dependency
[go-api](https://github.com/twreporter/go-api/pull/555/files)
[npm-packages](https://github.com/twreporter/twreporter-npm-packages/pull/410)

### DIFF
--- a/src/components/member-page/email-subscription/subscription-options.js
+++ b/src/components/member-page/email-subscription/subscription-options.js
@@ -99,9 +99,16 @@ const options = [
 const { actions, reduxStateFields } = twreporterRedux
 const { setUserData } = actions
 
-const SubscriptionOptions = ({ jwt, userID, setUserData }) => {
-  // TODO: get from state
-  const [newsletterSubscriptions, setNewsletterSubscriptions] = useState([])
+const SubscriptionOptions = ({
+  jwt,
+  userID,
+  setUserData,
+  readPreference,
+  maillist,
+}) => {
+  const [newsletterSubscriptions, setNewsletterSubscriptions] = useState([
+    ...maillist,
+  ])
 
   const [isToggleBtnDisabled, setIsToggleBtnDisabled] = useState(false)
   const { toastr } = useContext(EmailSubscriptionContext)
@@ -119,8 +126,7 @@ const SubscriptionOptions = ({ jwt, userID, setUserData }) => {
     }
     setNewsletterSubscriptions(subscriptions)
     setIsToggleBtnDisabled(true)
-    // TODO: get read_preference
-    setUserData(jwt, userID, [], subscriptions)
+    setUserData(jwt, userID, readPreference, subscriptions)
       .then(() => {
         setIsToggleBtnDisabled(false)
         toastr({ text: snackBarText })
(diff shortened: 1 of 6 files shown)


src/components/member-page/member-data.js
```diff
@@ -88,7 +85,7 @@ const MemberData = ({
 }
 
 MemberData.propTypes = {
-  role: PropTypes.oneOf(Object.values(MEMBER_ROLE)),
+  role: PropTypes.object,
```
Aylie-Chou, line 88: could we specify the shape of role object

liruchen32: fixed

src/containers/Member.js
```diff
@@ -130,7 +113,7 @@ const MemberPage = ({ releaseBranch = BRANCH.master }) => {
           <Route exact path={routes.memberPage.path}>
             <RoleCardContainer>
               <MemberRoleCard
-                role={memberData.role}
+                roleKey={memberData.role.key}
```
Aylie-Chou, line 116: Do we need to give default `memberData` value to ensure that `memberData.role` would not be undefined to prevent js error here ?

src/containers/Member.js
```diff
@@ -168,6 +151,27 @@ const MemberPage = ({ releaseBranch = BRANCH.master }) => {
 
 MemberPage.propTypes = {
   releaseBranch: BRANCH_PROP_TYPES,
+  memberData: propTypes.object,
```
Aylie-Chou, line 154: could we also specify the shape of memberData here

src/components/member-page/email-subscription/subscription-options.js
```diff
@@ -175,16 +181,26 @@ const SubscriptionOptions = ({ jwt, userID, setUserData }) => {
...
+    readPreference,
+    maillist,
   }
 }
 
 SubscriptionOptions.propTypes = {
   jwt: PropTypes.string,
   userID: PropTypes.number,
   setUserData: PropTypes.func.isRequired,
+  readPreference: PropTypes.array,
```
Aylie-Chou, line 202: It seems like this page would not use `readPreference` value.
Could we remove it for now ?

Aylie-Chou (changes requested):
Great work to add new `role.key` column & migrate to use it as the key for assets !
Please have a look at comments & Do we need to add loading effect when getting user data in member page ?

src/components/member-page/email-subscription/subscription-options.js
liruchen32: WE NEED THIS

src/containers/Member.js
liruchen32: fixed

src/components/member-page/email-subscription/subscription-options.js
Aylie-Chou: Is it because if we give null parameter to set user data api, it would clear user's readPreference value ?

Aylie-Chou (approved):
LGTM

PR #2666, opened 2023-11-09
feat: add article card skeleton & use load more in latest

Aylie-Chou:
# Issue
[[維運] 最新頁文章載入換成 load more button](https://app.asana.com/0/1203699264568808/1205754859454994/f)

# Notice
- add `Skeleton` component for article cards with RWD
- use load more instead of pagination
  - remove support of changing page with url search param `page`

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #2713, opened 2023-12-28
fix(src/containers/article): add beforeunload eventListener

liruchen32:
# Issue
[[會員制] 閱讀文章篇數、閱讀時間串接](https://app.asana.com/0/1203699264568808/1206008027632075/f) defects

# Notice

# Dependency

### DIFF
--- a/src/containers/Article.js
+++ b/src/containers/Article.js
@@ -220,7 +220,7 @@ class Article extends PureComponent {
       ) {
         this.sendActiveTime()
       }
-      return false
+      return true
     })
     if (prevProps.slugToFetch !== this.props.slugToFetch) {
       this.fetchAFullPostWithCatch(this.props.slugToFetch)
--- a/src/containers/Article.js
+++ b/src/containers/Article.js
@@ -85,7 +85,7 @@ class Article extends PureComponent {
       this._handleUserActivity.bind(this),
       500
     )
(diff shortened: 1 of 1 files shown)


src/containers/Article.js
```diff
@@ -160,25 +165,33 @@ class Article extends PureComponent {
...
+    }
+  }
+
   componentDidMount() {
-    const { fontLevel, changeFontLevel, slugToFetch, isAuthed } = this.props
+    const { fontLevel, changeFontLevel, slugToFetch } = this.props
 
     // Fetch the full post
     this.fetchAFullPostWithCatch(slugToFetch)
-    if (isAuthed) {
```
Aylie-Chou, line 168: Question: do we have any other method to register those event handler only when user has authed ?

Aylie-Chou (changes requested):
Good replacement to use `pagehide` instead of `beforeunload`
Just have the concern about the extra loading caused by new added event handler when user is not authed.
Please have a look ~

src/containers/Article.js
liruchen32: 有 debounce & throttle 應該不會造成太大的負擔

Aylie-Chou (approved):
LGTM
The reason of how `isAuthed` influence event register could be found out in following survey.

PR #2728, opened 2024-01-16
fix: set isActive to false after visibility change to hidden

liruchen32:
# Issue
set isActive to false after visibility change to hidden

# Notice

# Dependency


Aylie-Chou (approved):
ＬＧＴＭ

PR #2748, opened 2024-01-30
fix: check visibility at componentDidMount

liruchen32:
# Issue
timer won't stop if tab been hidden before componentDidMount

# Notice

# Dependency


Aylie-Chou (approved):
ＬＧＴＭ

PR #2763, opened 2024-02-15
fix: saved bookmark section loading & text

liruchen32:
# Issue
need to show empty page instead of loading spinner

# Notice

# Dependency


Aylie-Chou (approved):
LGTM
Could we also add the resolving issue link or description in PR conversations 🙏

PR #2775, opened 2024-02-21
fix: disable bookmark toggle in latest page

Aylie-Chou:
# Issue
disable bookmark feature before all defects resolved

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #2786, opened 2024-03-05
fix: replace state and give startReadingTime default Date.now

liruchen32:
# Issue
The default value of `startReadingTime` being 0 will cause inaccuracies in the calculation of reading time.

# Notice

# Dependency


nickhsine:
LGTM 👍

PR #2813, opened 2024-03-28
feat: article page refactor to functional componet

liruchen32:
# Issue
Refactor the article page into a functional component to see if it improves prevention of reading time errors.

# Notice

# Dependency

### DIFF
--- a/src/containers/Article-new.js
+++ b/src/containers/Article-new.js
@@ -1,0 +1,617 @@
+import React, { useState, useEffect, useRef } from 'react'
+import { connect } from 'react-redux'
+import { Link, withRouter } from 'react-router-dom'
+import PropTypes from 'prop-types'
+import Helmet from 'react-helmet'
+import TagManager from 'react-gtm-module'
+import localForage from 'localforage'
+import memoizeOne from 'memoize-one'
+// logger
+import loggerFactory from '../logger'
+// uiManager
+import uiManager from '../managers/ui-manager'
+// constants
+import bsConst from '../constants/browser-storage'
+import siteMeta from '../constants/site-meta'
+// utils
+import cloneUtils from '../utils/shallow-clone-entity'
+// components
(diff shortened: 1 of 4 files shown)


src/containers/Article-new.js
```diff
@@ -0,0 +1,617 @@
...
+      document.removeEventListener('scroll', handleScroll)
+      document.removeEventListener('visibilitychange', handleVisibilityChange)
+      document.removeEventListener('mousemove', handleUserActivity)
+      document.removeEventListener('click', handleUserActivity)
+      window.removeEventListener('pagehide', handlePagehide)
+    }
+  }, [])
+
+  useEffect(() => {
+    if (!isFetchingPost) {
```
Aylie-Chou, line 298: The default value of `isFetchingPost` is false, which means it's value might change like : false -> true -> false
Do we need to check the previous props value in order to match the false after true ?

src/containers/Article-new.js
```diff
@@ -0,0 +1,617 @@
...
+  reading_height: 0.75,
+  reading_time: 10000, // 10 second
+}
+const articleReadTimeConditionConfig = {
+  inactive_time: 900000, // 900 second
+  min_active_time: 5000, // 5 second
+}
+
+const Article = ({
+  slugToFetch,
```
Aylie-Chou, line 54: Nice to have:
Could we arrange the props to show that which is passed by redux state (mapStateByProps) & by parent

Aylie-Chou (changes requested):
Nice work to refactor article page !
This page is important for our users

Please have a look at comments ~~

src/containers/Article-new.js
liruchen32: fixed

src/containers/Article-new.js
liruchen32: yes sir

Aylie-Chou (approved):
LGTM

PR #2820, opened 2024-04-10
fix: get slug from post for canonical

liruchen32:
# Issue
Article page canonical tag got empty slug

# Notice

# Dependency


Aylie-Chou (approved):
LGTM

PR #2847, opened 2024-05-08
fix: build fail in node v18

Aylie-Chou:
# Issue
https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported

# Dependency
N/A


liruchen32 (approved):
LGTM

PR #2852, opened 2024-05-10
fix: fetch at mount while cause duplicated related articles

liruchen32:
# Issue
Duplicated related articles.

# Notice

# Dependency


Aylie-Chou (approved):
LGTM

PR #2885, opened 2024-05-23
fix: set isLoading to false after dispatch

liruchen32:
# Issue
[[defect]重要報導回顧-無資料時應正常顯示empty state](https://app.asana.com/0/1203699264568808/1207367216972411/f)

# Notice

# Dependency

### DIFF
--- a/src/components/my-reading/reviewing-article-section.js
+++ b/src/components/my-reading/reviewing-article-section.js
@@ -75,7 +75,6 @@ const ReviewingArticleSection = () => {
   }
 
   useEffect(() => {
-    setIsLoading(true)
     const jwt = _.get(state, [reduxStateFields.auth, 'accessToken'])
     dispatch(getPostReviews(jwt)).then(res => {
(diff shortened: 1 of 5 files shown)


src/components/my-reading/reviewing-article-section.js
```diff
@@ -103,8 +102,8 @@ const ReviewingArticleSection = () => {
       } else {
         setShowingReviewingArticle(formattedReviews)
       }
-      setIsLoading(false)
     })
+    setIsLoading(false)
```
Aylie-Chou, line 106: Question: do we need to wait for fetching data done (finally) before setting `isLoading` false ?

Aylie-Chou (changes requested):
as comment ~

src/components/my-reading/reviewing-article-section.js
liruchen32: fixed

Aylie-Chou (approved):
LGTM

PR #2948, opened 2024-07-18
fix: remove dompurify for search

liruchen32:
# Issue
Dompurify will cause search page crashed

# Notice

# Dependency


Aylie-Chou (approved):
ＬＧＴＭ
could we also memo this issue in notion doc ?

PR #2982, opened 2024-09-18
fix: change editor row count

liruchen32:
# Issue
[[spec change]有三個](https://app.asana.com/0/1203699264568808/1208338689540179/f)

# Notice

# Dependency


Aylie-Chou (approved):
ＬＧＴＭ

PR #3019, opened 2024-11-18
fix: use createSelector to avoid rerender

liruchen32:
# Issue
[文章頁｜header 收合/展開效果異常](https://app.asana.com/0/1203699264568808/1208761465986478/f)

# Notice
from 官網範例
```javascript
function TodoList() {
  // ❌ WARNING: this _always_ returns a new reference, so it will _always_ re-render!
  const completedTodos = useSelector(state =>
    state.todos.map(todo => todo.completed)
  )
}
```
[ref](https://redux.js.org/usage/deriving-data-selectors#optimizing-selectors-with-memoization)

# Dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 5.2.2-rc.2, 2024-11-18
+
+### Notable Changes
+
+- fix
+  - use createSelector to avoid rerender
+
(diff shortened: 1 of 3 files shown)


Aylie-Chou (approved):
LGTM👍
Would we need to check previous PRs to use memorized selector as well ?

PR #3042, opened 2024-12-12
fix: add material symbol font prefetch

Aylie-Chou:
# Issue
[[維運] 手機 icon 未即時顯示問題](https://www.notion.so/twreporter/icon-33860a6d7a0c4e59973ea716e01cbbbb?pvs=4)

# Notice
- develop environment serves material symbol fonts on webpack server which has different port, thus only prefetch the font on staging, production

# Dependency
https://github.com/twreporter/twreporter-npm-packages/pull/598


liruchen32 (approved):
LGTM

PR #3050, opened 2024-12-19
fix: join font faces with no comma

liruchen32:
# Issue
[[維運] 全站字體修復](https://www.notion.so/twreporter/3245d8e64b504944937681b020105401?pvs=4)

# Notice

# Dependency


Aylie-Chou (approved):
ＬＧＴＭ
could we also add the error cases as reference in this PR ?

PR #3106, opened 2025-06-03
fix: add www for url

liruchen32:
# Issue
will get cors error if url not include 'www'

# Notice

# Dependency


Aylie-Chou (approved):
LGTMwww

PR #3117, opened 2025-07-17
fix: handle offline donation at empty state

liruchen32:
# Issue
[[defect] 無贊助紀錄者不能正常顯示匯入提示popup](https://www.notion.so/twreporter/defect-popup-2328dbdca932804682c3c460e0f72f84?source=copy_link)

# Notice

# Dependency

### DIFF
--- a/src/components/member-page/donation/empty-donation.js
+++ b/src/components/member-page/donation/empty-donation.js
@@ -1,5 +1,6 @@
 import React, { useContext } from 'react'
 import styled from 'styled-components'
+import PropTypes from 'prop-types'
 // context
 import { CoreContext } from '../../../contexts'
 // @twreporter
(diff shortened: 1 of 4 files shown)


src/components/member-page/donation/empty-donation.js
```diff
@@ -42,13 +43,17 @@ export const EmptyDonation = () => {
...
           type={PillButton.Type.SECONDARY}
+          onClick={handleImportOfflineDonation}
         />,
       ]}
     />
   )
 }
+
+EmptyDonation.propTypes = {
+  handleImportOfflineDonation: PropTypes.func,
```
Aylie-Chou, line 58: Question: is handleImportOfflineDonation fun required ?

Aylie-Chou (commented):
Just want to check if the new added props should be required or not

